PR #26, opened 2020-08-10
cli: fix get-record by doi

### LINKED ISSUE #18: search by DOI not working
This works:

```console
$ cernopendata-client get-record --recid 4910
```

This does not:

```console
$ cernopendata-client get-record --doi '10.7483/OPENDATA.LHCB.N75T.TJPE'
Traceback (most recent call last):
  File "/home/simko/.virtualenvs/cernopendata/bin/cernopendata-client", line 8, in <module>
    sys.exit(cernopendata_client())
  File "/home/simko/.virtualenvs/cernopendata/lib/python3.8/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/home/simko/.virtualenvs/cernopendata/lib/python3.8/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/home/simko/.virtualenvs/cernopendata/lib/python3.8/site-packages/click/core.py", line 1135, in invoke
    sub_ctx = cmd.make_context(cmd_name, args, parent=ctx)
  File "/home/simko/.virtualenvs/cernopendata/lib/python3.8/site-packages/click/core.py", line 641, in make_context
    self.parse_args(ctx, args)
  File "/home/simko/.virtualenvs/cernopendata/lib/python3.8/site-packages/click/core.py", line 940, in parse_args
    value, args = param.handle_parse_result(ctx, opts, args)
  File "/home/simko/.virtualenvs/cernopendata/lib/python3.8/site-packages/click/core.py", line 1476, in handle_parse_result
    value = invoke_param_callback(
  File "/home/simko/.virtualenvs/cernopendata/lib/python3.8/site-packages/click/core.py", line 96, in invoke_param_callback
    return callback(ctx, param, value)
  File "/home/simko/.virtualenvs/cernopendata/lib/python3.8/site-packages/cernopendata_client/cli.py", line 31, in ensure_positive_int
    if value < 0:
TypeError: '<' not supported between instances of 'NoneType' and 'int'
```

To be fixed.

ParthS007:
Closes #18

### DIFF
--- a/cernopendata_client/cli.py
+++ b/cernopendata_client/cli.py
@@ -22,18 +22,15 @@ except ImportError:
     # fallback for Python 2
     from urllib import quote
 
-
-
 SEARCH_URL = 'http://opendata.cern.ch/api/records/'
 
 
-def ensure_positive_int(ctx, param, value):
-    if value < 0:
-        click.secho("Recid should be a positive integer", fg='red', err=True)
-        sys.exit()
-    else:
-        return value
-
+def validate_recid(recid):
+    if recid <= 0:
+        raise click.BadParameter(
+            'Recid should be a positive integer',
+            param_hint=["--recid"],
+        )
 
 @click.group()
 def cernopendata_client():
@@ -43,7 +40,6 @@ def cernopendata_client():
 @cernopendata_client.command()
 @click.option('--recid',
               type=click.INT,
-              callback=ensure_positive_int,
               help='Record ID')
 @click.option('--doi',
               help='Digital Object Identifier.')
@@ -55,6 +51,8 @@ def cernopendata_client():
 def get_record(recid, doi, title, output_fields):
     """Get records content by its recid, doi or title filtered."""
     # TODO: Add decorator to require one of (recid, doi or title)
+    if recid is not None:
+        validate_recid(recid)
     record_json = get_record_as_json(recid, doi, title)
     if output_fields is not None:
         output_fields = [f.strip() for f in output_fields.split(',')]
@@ -79,9 +77,8 @@ def get_record(recid, doi, title, output_fields):
 
 @cernopendata_client.command()
 @click.option('--recid',
-              help='Record ID',
               type=click.INT,
-              callback=ensure_positive_int)
+              help='Record ID')
 @click.option('--doi',
               help='Digital Object Identifier.')
 @click.option('--title',
@@ -93,6 +90,8 @@ def get_record(recid, doi, title, output_fields):
               help='Expand file indexes? [default=yes]')
 def get_file_locations(recid, doi, title, protocol, expand):
     """Get a list of files belonging to a dataset."""
+    if recid is not None:
+        validate_recid(recid)
     record_json = get_record_as_json(recid, doi, title)
     file_locations = [
         file['uri'] for file in record_json['metadata']['files']
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -7,4 +7,5 @@ The list of contributors in alphabetical order:
 - `Alex Mieth <https://github.com/AlexMieth>`_
 - `Daniel Prelipcean <https://orcid.org/0000-0002-4855-194X>`_
 - `Jan Okraska <https://orcid.org/0000-0002-1416-3244>`_
+- `Parth Shandilya <https://github.com/ParthS007>`_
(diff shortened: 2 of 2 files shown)


cernopendata_client/cli.py
```diff
@@ -26,14 +26,19 @@
 
 SEARCH_URL = 'http://opendata.cern.ch/api/records/'
 
-
+# Callback
 def ensure_positive_int(ctx, param, value):
```
tiborsimko, line 30: I think the callback can be safely removed now?

cernopendata_client/cli.py
```diff
@@ -26,14 +26,19 @@
...
-    else:
+        if value < 0:
+            click.secho("Recid should be a positive integer", fg='red', err=True)
+            sys.exit()
         return value
 
+# Custom Type
+class ArgumentCheck(click.ParamType):
+    def ensure_positive_int(self, value, param, ctx):
+        if value < 0:
```
tiborsimko, line 39: Please move "Custom Type" comment into a class docstring explaining what the class does.

cernopendata_client/cli.py
```diff
@@ -55,6 +59,11 @@ def cernopendata_client():
 def get_record(recid, doi, title, output_fields):
     """Get records content by its recid, doi or title filtered."""
     # TODO: Add decorator to require one of (recid, doi or title)
+    if recid is not None and recid < 0: # This works
```
tiborsimko, line 62: Works well indeed :+1: except for the recid == 0 situation where one gets:

```console
$ cernopendata-client get-record --recid 0
Please provide at least one of following arguments: (recid, doi, title)
```

One should get the usual message:

```console
$ $ cernopendata-client get-record --recid -1
Usage: cernopendata-client get-record [OPTIONS]

Error: Invalid value for '--recid': Recid should be a positive integer
```

tiborsimko:
Commit cosmetics: Please add `Closes #18` also to the commit log message, and add yourself to the `AUTHORS.rst` file.

cernopendata_client/cli.py
```diff
@@ -55,6 +43,11 @@ def cernopendata_client():
 def get_record(recid, doi, title, output_fields):
     """Get records content by its recid, doi or title filtered."""
     # TODO: Add decorator to require one of (recid, doi or title)
+    if recid is not None and recid <= 0:
+        raise click.BadParameter(
+            'Recid should be a positive integer',
+            param_hint=["--recid"],
+        )
```
tiborsimko, line 50: Note that you have to return and not continue, otherwise some bad URLs are tried:

```console
$ cernopendata-client get-file-locations --recid -1
ERROR: The record id number you supplied is not valid.
404 Client Error: NOT FOUND for url: http://opendata.cern.ch/record/-1
$ cernopendata-client get-file-locations --recid hello
ERROR: The record id number you supplied is not valid.
404 Client Error: NOT FOUND for url: http://opendata.cern.ch/record/hello
```

ParthS007: It is returning but I have to add the validation in `get_file_location` method as well.

cernopendata_client/cli.py
```diff
@@ -93,6 +85,11 @@ def get_record(recid, doi, title, output_fields):
               help='Expand file indexes? [default=yes]')
 def get_file_locations(recid, doi, title, protocol, expand):
     """Get a list of files belonging to a dataset."""
+    if recid is not None and recid <= 0:
```
ParthS007, line 88: This validation is coming two times, Should I go ahead with creating a sepearte function and calling it or this is fine?

tiborsimko: Yes, we can refactor everything... There will be many more functions in the future, so it will be useful to take the validation out. 

Speaking pseudo-code style, we currently have three ways how to select records: via recid, via doi, via title.  These can maybe later accept wildcards so that the user can comfortably select more than one related dataset (e.g. `title='/BTag/*'`). Imagine we have `get-record` now, but we add a new `get-records` in the future. So why don't we introduce a "validate" function that we could plug everywhere where it is needed, but also another "select" function that will return list of record IDs based by matching recid, or doi, or title, and the caller can decide what to do with the findings, e.g. if somebody asks for one record in `get-record` we first run the validator function to make sure arguments are OK, then the record selector function to find matching record IDs, and if the selector finds more than one record, then we should abort with an error, and if it finds one record, we get JSON for it by hitting REST API.

PR #32, opened 2020-09-02
Refactoring of CLI and Configure alternative server URL

### LINKED ISSUE #20: deprecate opendata_analysis_query.py
`/opendata_analysis_query.py` contains old code that is not used anymore.

Check functionality and port necessary one to the new structure.

Remove the files and its documentation afterwards.

### LINKED ISSUE #31: configuration: allow setting alternative server URL
Currently the `cernopendata-client` is querying `opendata.cern.ch` instance, and this is hard-coded:

```
$ rg opendata.cern.ch
cernopendata_client/cli.py
24:SEARCH_URL = "http://opendata.cern.ch/api/records/"
103:                    "root://eospublic.cern.ch/", "http://opendata.cern.ch"
117:            f.replace("root://eospublic.cern.ch/", "http://opendata.cern.ch")
```

It would be good to allow querying `opendata-qa.cern.ch` or `opendata-dev.cern.ch` instances, both for testing purposes and when working with open data releases that take a long time to "mature" on the DEV or QA instances before making it to the PROD instance.

We could solve it by adding a new command-line option (`--server <url>`) to each CLI command, where users could pass a wanted server instance, for example:

```console
$ cernopendata-client get-file-locations --recid 15007 --protocol http --server http://opendata-dev.cern.ch
```

 If the option is not used, the client would connect by default to `http://opendata.cern.ch`.

ParthS007:
closes #31 
closes #20


.gitignore
```diff
@@ -89,6 +89,7 @@ venv/
 ENV/
 env.bak/
 venv.bak/
+.vscode/
```
tiborsimko, line 92: Note that you can put this also to your personal global git ignore, so that it would apply on all your git projects.

cernopendata_client/cli.py
```diff
@@ -40,20 +30,26 @@ def cernopendata_client():
 @cernopendata_client.command()
 @click.option("--recid", type=click.INT, help="Record ID")
 @click.option("--doi", help="Digital Object Identifier.")
-@click.option("--title", help="Record title")
+@click.option("--title", help="Record Title")
```
tiborsimko, line 33: `Record title` is OK, we mostly use British English where titles are not usually that capitalised as in US English...

That said, we should amend help texts later, to provide nicer examples for people reading help pages like https://cernopendata-client.readthedocs.io/en/latest/cliapi.html#cernopendata-client-get-file-locations  I can expand that later.

cernopendata_client/cli.py
```diff
@@ -40,20 +30,26 @@ def cernopendata_client():
...
+    type=click.STRING,
+    help="Query Server URL",
+)
+def get_record(recid, doi, title, output_fields, server):
     """Get records content by its recid, doi or title filtered."""
     # TODO: Add decorator to require one of (recid, doi or title)
     if recid is not None:
         validate_recid(recid)
-    record_json = get_record_as_json(recid, doi, title)
+    record_json = get_record_as_json(recid, doi, title, server)
```
tiborsimko, line 52: Please use `server` as the first parameter to all functions.  It seems more intuitive to write `get_value(server, key)` i.e. first saying from which server we are fetching information, and only then saying the inputs.

cernopendata_client/cli.py
```diff
@@ -88,20 +84,24 @@ def get_record(recid, doi, title, output_fields):
 @click.option(
     "--expand/--no-expand", default=True, help="Expand file indexes? [default=yes]"
 )
-def get_file_locations(recid, doi, title, protocol, expand):
+@click.option(
+    "--server",
+    default="http://opendata.cern.ch",
+    type=click.STRING,
+    help="Query server URL",
```
tiborsimko, line 91: The help string can say more explicitly "Which CERN Open Data server to query? [default=http://opendata.cern.ch]".  Please check also how help looks like when running `--help`.

cernopendata_client/cli.py
```diff
@@ -88,20 +84,24 @@ def get_record(recid, doi, title, output_fields):
...
+)
+def get_file_locations(recid, doi, title, protocol, expand, server):
```
tiborsimko, line 93: `server` argument first again.

cernopendata_client/opendata_analysis_query.py
```diff
@@ -1,208 +0,0 @@
-from __future__ import print_function
```
tiborsimko, line 1: Are we ready to delete this file already?  If so, you can close issue #20 as well.

cernopendata_client/search.py
```diff
@@ -0,0 +1,110 @@
...
+        else:
+            try:
+                input_record_url_check.raise_for_status()
+            except requests.HTTPError as http_error_msg:
+                print("ERROR: The record id number you supplied is not valid.")
+                sys.exit(http_error_msg)
+            return False
+
+
+def get_recid_api(base_record_id=None, server=None):
```
tiborsimko, line 47: `server` first, as elsewhere.

Please also think of providing some validation.  Currently one gets:

```console
$ cernopendata-client get-file-locations --recid 15007 --protocol http --server something 
...
  File "/home/simko/.virtualenvs/cernopendata-client/lib/python3.8/site-packages/requests/models.py", line 387, in prepare_url
    raise MissingSchema(error)
requests.exceptions.MissingSchema: Invalid URL 'something/record/15007': No schema supplied. Perhaps you meant http://something/record/15007?
```

This is good, but seeing Python traceback may be confusing for some users.

Perhaps validate server arguments after they are passed by:
- adding a validator function in `validator.py`
- calling it right after click entry points
- the validator could for now simply check whether the argument starts with `http://...` and return true/false
- in case of wrong argument, we can fail gracefully with an error message, not showing python traceback

tiborsimko (commented):
Works well, left a few comments about putting `server` as first parameter to all functions, and about validating the passed value so that users won't see so much Python traceback in case of wrong user input.

cernopendata_client/opendata_analysis_query.py
ParthS007: Yes, I think so. I have moved the used two functions in search.py and other functions are not used anywhere.

tiborsimko (approved):

PR #34, opened 2020-09-04
cli: download files feature

### LINKED ISSUE #22: download-files: initial release
If a user wishes to download files belonging to a record, the current technique is to list file locations:

```console 
$ cernopendata-client get-file-locations --recid 5500 --protocol http
http://opendata.cern.ch/eos/opendata/cms/software/HiggsExample20112012/BuildFile.xml
http://opendata.cern.ch/eos/opendata/cms/software/HiggsExample20112012/HiggsDemoAnalyzer.cc
http://opendata.cern.ch/eos/opendata/cms/software/HiggsExample20112012/List_indexfile.txt
http://opendata.cern.ch/eos/opendata/cms/software/HiggsExample20112012/M4Lnormdatall.cc
http://opendata.cern.ch/eos/opendata/cms/software/HiggsExample20112012/M4Lnormdatall_lvl3.cc
http://opendata.cern.ch/eos/opendata/cms/software/HiggsExample20112012/demoanalyzer_cfg_level3MC.py
http://opendata.cern.ch/eos/opendata/cms/software/HiggsExample20112012/demoanalyzer_cfg_level3data.py
http://opendata.cern.ch/eos/opendata/cms/software/HiggsExample20112012/demoanalyzer_cfg_level4MC.py
http://opendata.cern.ch/eos/opendata/cms/software/HiggsExample20112012/demoanalyzer_cfg_level4data.py
http://opendata.cern.ch/eos/opendata/cms/software/HiggsExample20112012/mass4l_combine.pdf
http://opendata.cern.ch/eos/opendata/cms/software/HiggsExample20112012/mass4l_combine.png
```

and then launch `wget` or `curl` commands to download them.

The goal of this issue is to simplify this process by introducing new command `download-files` that would do this for the user.

Possible options:

```console
$ cernopendata-client download-files --recid 5500 --protocol http --parallel-processes 2
```

This would launch two parallel downloading processes, using a suitable Python library, to download the files into current directory.

P.S.: MVP is simply to download files; resuming interrupted downloads will be part of another issue, but it is good to think about this functionality upfront.

P.S. An option `--target-directory` could be introduced which would recreate directory structure known from the original record.  This will be important for AOD files which have subdirectory structure such as [this one](https://eospublichttp.cern.ch/eos/opendata/cms/Run2010A/BTau/AOD/Apr21ReReco-v1).  So the corresponding subdirectories would have to be created in the target directory.

ParthS007:
closes #22

### DIFF
--- a/cernopendata_client/cli.py
+++ b/cernopendata_client/cli.py
@@ -7,17 +7,19 @@
 # it under the terms of the GPLv3 license; see LICENSE file for more details.
 
 import json
-import sys
+import os, sys
 
 import click
 import requests
-from requests import HTTPError
+import pycurl
 
+from cernopendata_client.tui import show_download_progress
 from cernopendata_client.search import (
     get_recid,
     verify_recid,
     get_recid_api,
     get_record_as_json,
+    get_files_list,
 )
 from cernopendata_client.validator import (
     validate_recid,
@@ -81,8 +83,8 @@ def get_record(server, recid, doi, title, output_fields):
 @click.option("--title", help="Record title")
 @click.option(
     "--protocol",
-    default="root",
-    type=click.Choice(["root", "http"]),
+    default="http",
+    type=click.Choice(["http", "root"]),
     help="Protocol to be used in links.",
 )
 @click.option(
@@ -100,25 +102,68 @@ def get_file_locations(server, recid, doi, title, protocol, expand):
     if recid is not None:
         validate_recid(recid)
     record_json = get_record_as_json(server, recid, doi, title)
-    file_locations = [file["uri"] for file in record_json["metadata"]["files"]]
-    if expand:
-        # let's unwind file indexes
-        file_locations_expanded = []
-        for file in file_locations:
-            if file.endswith("_file_index.txt"):
-                url_file = file.replace("root://eospublic.cern.ch/", server)
-                req = requests.get(url_file)
-                for url_individual_file in req.text.split("\n"):
-                    if url_individual_file:
-                        file_locations_expanded.append(url_individual_file)
-            elif file.endswith("_file_index.json"):
-                pass
-            else:
-                file_locations_expanded.append(file)
-        file_locations = file_locations_expanded
-
-    if protocol == "http":
-        file_locations = [
-            f.replace("root://eospublic.cern.ch/", server) for f in file_locations
-        ]
+    file_locations = get_files_list(server, record_json, protocol, expand)
     click.echo("\n".join(file_locations))
+
+
+@cernopendata_client.command()
+@click.option("--recid", type=click.INT, help="Record ID")
+@click.option("--doi", help="Digital Object Identifier.")
+@click.option("--title", help="Record title")
+@click.option(
+    "--protocol",
+    default="http",
+    type=click.Choice(["http", "root"]),
+    help="Protocol to be used in links.",
+)
+@click.option(
+    "--expand/--no-expand", default=True, help="Expand file indexes? [default=yes]"
+)
+@click.option(
+    "--server",
+    default="http://opendata.cern.ch",
+    type=click.STRING,
+    help="Which CERN Open Data server to query? [default=http://opendata.cern.ch]",
+)
+def download_files(server, recid, doi, title, protocol, expand):
+    """Download the list of files belonging to a dataset."""
+    validate_server(server)
+    if recid is not None:
+        validate_recid(recid)
+    if protocol == "root":
+        click.secho(
+            "Root protocol is not supported yet.",
+            fg="red",
+            err=True,
+        )
+        sys.exit(1)
+    record_json = get_record_as_json(server, recid, doi, title)
+    file_locations = get_files_list(server, record_json, protocol, expand)
+
+    path = str(recid)
+    if not os.path.isdir(path):
+        try:
+            os.mkdir(path)
+        except OSError:
+            print("Creation of the directory %s failed" % path)
+    total_files = len(file_locations)
+    for file_location in file_locations:
+        file_name = file_location.split("/")[-1]
+        file_dest = path + "/" + file_name
+        with open(file_dest, "wb") as f:
+            print(
+                "==> Downloading file {} of {}: ./{}/{}".format(
+                    file_locations.index(file_location) + 1,
+                    total_files,
+                    path,
+                    file_name,
+                )
+            )
+            c = pycurl.Curl()
+            c.setopt(c.URL, file_location)
+            c.setopt(c.WRITEDATA, f)
+            c.setopt(c.NOPROGRESS, False)
+            c.setopt(c.XFERINFOFUNCTION, show_download_progress)
+            c.perform()
+            c.close()
+    click.echo("\nDownload completed!")
--- a/setup.py
+++ b/setup.py
@@ -44,7 +44,7 @@ setup_requires = [
     "pytest-runner>=2.7",
 ]
 
-install_requires = ["click>=7,<8", "requests==2.22.0"]
+install_requires = ["click>=7,<8", "requests==2.22.0", "pycurl>=7"]
 
 # Get the version string. Cannot be done with import!
 with open(os.path.join("cernopendata_client", "version.py"), "rt") as f:
--- a/tests/tests_cli/test_download_files.py
+++ b/tests/tests_cli/test_download_files.py
@@ -1,0 +1,29 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of cernopendata-client.
+#
+# Copyright (C) 2020 CERN.
+#
+# cernopendata-client is free software; you can redistribute it and/or modify
+# it under the terms of the GPLv3 license; see LICENSE file for more details.
+
+"""cernopendata-client cli command download-files test."""
+
+import os
+from click.testing import CliRunner
+from cernopendata_client.cli import download_files
+
+
+def test_download_files():
+    """Test download_files() command."""
+    test_file = "3005/0d0714743f0204ed3c0144941e6ce248.configFile.py"
+    if os.path.isfile(test_file):
+        os.remove(test_file)
+    test_download_files = CliRunner()
+    test_result = test_download_files.invoke(download_files, ["--recid", 3005])
+    assert test_result.exit_code == 0
+    assert os.path.isfile(test_file) is True
+    assert os.path.getsize(test_file) == 3644
+    assert test_result.output.endswith("\nDownload completed!\n")
+    if os.path.isfile(test_file):
+        os.remove(test_file)
--- a/tests/tests_cli/test_get_file_locations.py
+++ b/tests/tests_cli/test_get_file_locations.py
@@ -1,0 +1,22 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of cernopendata-client.
+#
+# Copyright (C) 2020 CERN.
+#
+# cernopendata-client is free software; you can redistribute it and/or modify
+# it under the terms of the GPLv3 license; see LICENSE file for more details.
+
+"""cernopendata-client cli command get-file-locations test."""
+
+from click.testing import CliRunner
+from cernopendata_client.cli import get_file_locations
+
+
+def test_get_file_locations():
+    """Test get-file-locations command."""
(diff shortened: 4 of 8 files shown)


cernopendata_client/cli.py
```diff
@@ -100,25 +102,51 @@ def get_file_locations(server, recid, doi, title, protocol, expand):
...
-        ]
+    file_locations = get_files_list(server, record_json, protocol, expand)
     click.echo("\n".join(file_locations))
+
+
+@cernopendata_client.command()
+@click.option("--recid", type=click.INT, help="Record ID")
+@click.option("--doi", help="Digital Object Identifier.")
+@click.option("--title", help="Record title")
+@click.option(
```
tiborsimko, line 113: Since we don't have `root` supported yet, maybe don't introduce this option... or, better, switch the defaults to use `http` and if people select `root` then reply that this is not yet supported and do `sys.exit(1)`.

cernopendata_client/cli.py
```diff
@@ -100,25 +102,51 @@ def get_file_locations(server, recid, doi, title, protocol, expand):
...
+    path = str(recid)
+    try:
+        os.mkdir(path)
+    except OSError:
+        print("Creation of the directory %s failed" % path)
+
+    for file_location in file_locations:
+        file_dest = path + "/" + file_location.split("/")[-1]
+        with open(file_dest, "wb") as f:
+            print("File Downloading")
```
tiborsimko, line 145: It would be nice to introduce some output formatting cosmetics, e.g. currently people see:

```console
$ cernopendata-client download-files --recid 5500 --protocol http
Creation of the directory 5500 failed
File Downloading
File Downloading kiB (100%)
File Downloading81 kiB (100%)
File Downloading kiB (100%)
File Downloading14 kiB (100%)
File Downloading15 kiB (100%)
File Downloading kiB (100%)
File Downloading kiB (100%)
File Downloading kiB (100%)
File Downloading kiB (100%)
File Downloading17 kiB (100%)
Downloading: 90/90 kiB (100%)
```

I would expect to see the URI or at least the file name that is being currently downloaded.  Also, the total count, so that people can estimate the ETA.  Somethhing like this:

```
==> Downloading file 1 of 10:  demoanalyzer_cfg_level3data.py
Downloading 123/123456789 KiB (1.23%)
==> Downloading file 2 of 10: demoanalyzer_cfg_level3MC.py
...
```

setup.py
```diff
@@ -45,7 +45,7 @@
     "pytest-runner>=2.7",
 ]
 
-install_requires = ["click>=7,<8", "requests==2.22.0"]
+install_requires = ["click>=7,<8", "requests==2.22.0", "pycurl>=7"]
```
tiborsimko, line 48: Travis CI fails with:

```
src/pycurl.h:202:30: fatal error: gnutls/gnutls.h: No such file or directory
```

You may need to amend `.travis.yml` to preinstall `gnutls` libraries.

cernopendata_client/helper.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
```
tiborsimko, line 1: `helper.py` is a rather generic name... since it can help with anything.  Better invent some more targeted name. 

- One option is to put it in `download.py` and have all the functions related to downloads sitting in that file.
- Another option is to put it in `tui.py` and have all the "text user interface" functions sitting in that file.

The former is probably better, since we wouldn't be easily able to reuse this function elsewhere in other contexts than downloading.

WDYT?

cernopendata_client/cli.py
tiborsimko: (we could also with default to 'http' over 'root' for the other `get-file-locations` function)

cernopendata_client/helper.py
ParthS007: The reason why I choose `helper.py` is that I thought we will be adding other functions that are on their own and don't require any data apart from arguments and we can add other functions not just limited to download functionality in that file which are doing some calculations or providing customized output on the terminal.

I think `tui.py` will fulfill this purpose and more understandable than `helper.py` or `download.py`. 

going with `tui.py` then 👍

tiborsimko: I meant that if this function is not going to be called by anybody except the `download-...` set of commands, then `download.py` may be appropriate location. If it would be called by `download-...` and `get-...` etc commands, for example styling general output messages, then `tui.py` may be more appropriate location. I'm OK with both, as `cernopendata-client` will grow, we may actually be revamping things later. (E.g. `reana-dev` started as one single file, but as the number of commands grew, we separate each command set into its own file.)

ParthS007: I agree and it sounds perfect. 👍

cernopendata_client/cli.py
```diff
@@ -81,7 +83,7 @@ def get_record(server, recid, doi, title, output_fields):
 @click.option("--title", help="Record title")
 @click.option(
     "--protocol",
-    default="root",
+    default="http",
     type=click.Choice(["root", "http"]),
```
tiborsimko, line 87: Please order alphabetically, `type=click.Choice(["http", "root"]),` here and in the other function.

cernopendata_client/cli.py
```diff
@@ -100,25 +102,67 @@ def get_file_locations(server, recid, doi, title, protocol, expand):
...
+            fg="red",
+            err=True,
+        )
+        sys.exit(1)
+    record_json = get_record_as_json(server, recid, doi, title)
+    file_locations = get_files_list(server, record_json, protocol, expand)
+
+    path = str(recid)
+    try:
+        os.mkdir(path)
```
tiborsimko, line 145: Please check first whether the directory does not already exists; there is no need to create it then.  Users would see a message like `Creation of the directory 5500 failed` and might be confused.

tiborsimko (commented):
Works well :+1: I left two more cosmetic comments. 

@ParthS007 Shall we merge before adding tests? Or after?

ParthS007:
@tiborsimko  I can write the test for testing the download helper in `tui.py` now. Please provide me the location of the file we discussed in person.

tiborsimko:
> I can write the test for testing the download helper in `tui.py` now. Please provide me the location of the file we discussed in person.

Good then! The tests for `tui.py` should go to `tests/test_tui.py` file. You can emulate `test_validator.py` that contain tests for `validator.py`.

tiborsimko:
> > I can write the test for testing the download helper in `tui.py` now. Please provide me the location of the file we discussed in person.
> 
> Good then! The tests for `tui.py` should go to `tests/test_tui.py` file. You can emulate `test_validator.py` that contain tests for `validator.py`.

We were testing `http://opendata.cern.ch/record/1050/files/2010RunBlumi.txt` but record 1050 has three files; let's perhaps use another file which has only one?  Example: `http://opendata.cern.ch/record/3005/files/0d0714743f0204ed3c0144941e6ce248.configFile.py`

ParthS007:
Testing one file seems fine to me for now, once I implement a low-level download_function, then we can add other test cases as well. what do you think?

ParthS007:
@tiborsimko I added basic tests for testing CLI commands and since we are not calling directly `show_download_progress`, I thought of writing a test for `download-files` command instead of `show_download_progress`.

Is this the right way to approach this?

Please let me know, I will do the changes if required.

For `get_record` and `get_file_locations`, I added basic tests for checking the output type on the basis of `recid`.

tests/test_cli.py
```diff
@@ -0,0 +1,34 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of cernopendata-client.
+#
+# Copyright (C) 2020 CERN.
+#
+# cernopendata-client is free software; you can redistribute it and/or modify
+# it under the terms of the GPLv3 license; see LICENSE file for more details.
+
+"""cernopendata-client text user interface(tui) test."""
```
tiborsimko, line 10: This test suite will grow, so please create three test files, one for each CLI command:

- `tests/test_get_record.py`
- `tests/test_get_file_locations.py`
- `tests/test_download_files.py`

?

tests/test_cli.py
```diff
@@ -0,0 +1,34 @@
...
+
+from click.testing import CliRunner
+from cernopendata_client.cli import get_record, get_file_locations, download_files
+
+
+def test_get_record():
+    """Test get_record() command."""
+    test_get_record = CliRunner()
+    test_result = test_get_record.invoke(get_record, ["--recid", 3005])
+    assert type(test_result.output) == str
```
tiborsimko, line 20: Instead of just asserting type string, you could also assert exit code (should be 0) and, more importantly, the presence of some expected text in the output.

Example:

```python
    assert '"title": "Configuration file for LHE step HIG-Summer11pLHE-00114_1_cfg.py"' in test_result.output
    assert test_result.exit_code == 0
```

The same for the other commands, e.g. when listing file locations, assert presence of a known one what the command should return, etc.

tiborsimko (commented):
> Is this the right way to approach this?

Yes, I like it :+1: I left two comments about (i) splitting tests for each command into its own file, and about (ii) enriching asserts by looking for expected strings.

tests/test_cli.py
ParthS007: Creating three test files sounds good, I was thinking of one more folder like

- `tests/tests_cli/test_get_record.py`
- `tests/tests_cli/test_get_file_locations.py`
- `tests/tests_cli/test_download_files.py`

which one looks fine?

tiborsimko: Yes, that's also OK, especially if we are going to split `cli.py` into `cli/command1.py`, `cli/command2.py` one day...

tests/tests_cli/test_get_file_locations.py
```diff
@@ -0,0 +1,22 @@
...
+from cernopendata_client.cli import get_file_locations
+
+
+def test_get_file_locations():
+    """Test get-file-locations command."""
+    test_get_file_locations = CliRunner()
+    test_result = test_get_file_locations.invoke(get_file_locations, ["--recid", 3005])
+    assert test_result.exit_code == 0
+    assert type(test_result.output) == str
+    assert 'configFile.py' in test_result.output
```
tiborsimko, line 22: Better to test the full name `0d0714743f0204ed3c0144941e6ce248.configFile.py` so that we are sure we did not get a file from another record.  (There are many `xxxxx.configFile.py` ones.)

tests/tests_cli/test_download_files.py
```diff
@@ -0,0 +1,21 @@
...
+from click.testing import CliRunner
+from cernopendata_client.cli import download_files
+
+
+def test_download_files():
+    """Test download_files() command."""
+    test_download_files = CliRunner()
+    test_result = test_download_files.invoke(download_files, ["--recid", 3005])
+    assert test_result.exit_code == 0
+    assert test_result.output.endswith("\nDownload completed!\n")
```
tiborsimko, line 21: You can also test whether `3005/0d0714743f0204ed3c0144941e6ce248.configFile.py` file is well present and whether its size is 3644 bytes.

BTW it may be necessary before running this test case to delete that file if it is present... so that we don't get false positives...

tests/tests_cli/test_download_files.py
```diff
@@ -0,0 +1,27 @@
...
+    """Test download_files() command."""
+    test_file = "3005/0d0714743f0204ed3c0144941e6ce248.configFile.py"
+    if os.path.isfile(test_file):
+        os.remove(test_file)
+    test_download_files = CliRunner()
+    test_result = test_download_files.invoke(download_files, ["--recid", 3005])
+    assert test_result.exit_code == 0
+    assert os.path.isfile(test_file) is True
+    assert os.path.getsize(test_file) == 3644
+    assert test_result.output.endswith("\nDownload completed!\n")
```
tiborsimko, line 27: Nice, please also remove the file after test is done as the last step.

BTW I get `black' warnings on the two test files...

tiborsimko: Note also the first comment, please add:

```
    if os.path.isfile(test_file):
        os.remove(test_file)
```

as the last instructions in the `test_download_files()` function, so that we end up with a blank slate again.

After this is done, we are ready to go :tada:

tiborsimko (approved):

Issue #40, opened 2020-09-11
docs: ReadTheDocs builds fail due to pycurl


tiborsimko:
After addition of `pycurl`, the ReadTheDocs builds fail with:

```
   FileNotFoundError: [Errno 2] No such file or directory: 'curl-config': 'curl-config'
```

This is most probably because some system dependencies such as `libssl-dev` and friends are not installed.

Investigate a best solution to fix the documentation builds.  
- Perhaps add these libraries if RTD build system allows this?
- Perhaps use `autodoc_mock_imports`?

ParthS007:
@tiborsimko this can be closed now?

tiborsimko:
I thought of looking into generating the CLI API in other way... E.g. we could run click-sphinx locally and add the file into `docs` so that ReadTheDocs would simply consume it from there.  (A bit like we are doing with OpenAPI stuff in REANA.) Hence I used `addresses` and not `closes` yet... I'll move the issue into "Ready for work" kanban column.

tiborsimko:
Closing for v0.1.0 release, we can attack this later.

PR #43, opened 2020-09-11
docs: autodoc_mock_imports for pycurl

tiborsimko:
Adds mocking of `pycurl` that cannot be installed on ReadTheDocs
builders.  However, we would also have to remove `pycurl` as the
installation dependency of `cernopendata-client` in `setup.py`, which
is not wanted at this stage.

Hence amends CLI API page to simply instruct persons to use command
line to get the list of all commands and options. It is better to
show latest user guide instead.

This fixes the ReadTheDocs building problem.

Addresses #40.

Signed-off-by: Tibor Šimko <tibor.simko@cern.ch>

### DIFF
--- a/docs/cliapi.rst
+++ b/docs/cliapi.rst
@@ -3,6 +3,9 @@
 CLI API
 =======
 
+Please use ``cernopendata-client --help`` in your terminal shell to learn about
+all commands and their options.
+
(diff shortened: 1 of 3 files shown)


ParthS007 (approved):
LGTM 🚀

PR #44, opened 2020-09-14
cli: download-files --file option

### LINKED ISSUE #38: download-files: download particular files only
[Record 1](http://opendata.cern.ch/record/1) contains 2916 files and is 2.7 TB big.

Chances are people would like to download it in batches.

Currently, `cernopendata-client download-files` would download everything. We need to introduce finer granularity.

The goal of this issue is to introduce a new option for `download-files`, called perhaps `--file`, which would download only one particular file:

```console
$ cernopendata-client download-files --recid 1 --filename 105FD6D0-8B71-E011-9613-00E081791775.root
```

Alternatively, we could offer regexp-like matching:

```console
$ cernopendata-client download-files --recid 1 --filename '*E011*'
```

which would download all files matching the given glob expression.

Alternatively, since the file order is perfectly defined in JSON, we can offer downloading by chunks, either given file number N1, or several files from file number N2 to file number N3:

```console
$ cernopendata-client download-files --recid 1 --filenumber 13
$ cernopendata-client download-files --recid 1 --filenumber 20-29
$ cernopendata-client download-files --recid 1 --filenumber 30-39
... 
```

ParthS007:
closes #38
added support for downloading files by name, regexp, file range

Now working on adding tests. 

@tiborsimko What could be written in the help string for this `--file` option?


cernopendata_client/cli.py
```diff
@@ -152,7 +152,8 @@ def get_file_locations(server, recid, doi, title, protocol, expand):
     type=click.STRING,
     help="Which CERN Open Data server to query? [default=http://opendata.cern.ch]",
 )
-def download_files(server, recid, doi, title, protocol, expand):
+@click.option("--file", nargs=2, type=str, help="")
+def download_files(server, recid, doi, title, protocol, expand, file):
```
tiborsimko, line 156: - [ ] Can you please add help message to document the option?  
- [ ] Also, add some example below in the docstring, following the current ones.
- [ ] Also, update the `docs/usage.rst` user guide mentioning this possibility.
- [ ] Coverage decreased as part of this PR, which is OK, but if you separate filtering functions outside the main entry point function, then you could write test cases for those.
- [ ] Note that we may need to introduce several options, e.g. if somebody ask for `--filename '*.root'` there could be O(3000) files, so it makes sense to also offer navigation by file index number.  So either we have one option and call it filter:

```
--filter name '*.root' --filter range 10-19
```

or we have several options:

```
--filter-filename '*.root' --filter-files 10-19
``` 
```

File: cernopendata_client/cli.py [hunk 2]
```diff
@@ -181,24 +208,29 @@ def download_files(server, recid, doi, title, protocol, expand):
...
+
+    if name:
+        download_file_locations = get_download_files_by_name(
+            name=name, file_locations=file_locations
+        )
+    elif regexp:
+        download_file_locations = get_download_files_by_regexp(
+            regexp=regexp, file_locations=file_locations
+        )
+    elif range:
```
tiborsimko, line 220: `elif` makes filters mutually exclusive, so one cannot do:

```
$ cernopendata-client download-files --recid 5500 --filter-regexp py --filter-range 1-2
```

This would be useful.

Two solutions:
- document that they are exclusive;
- amend `elif` into a series of filters applied one after another  

In the latter case there is no need to cache anything on the server side, you can simply apply first filter to get the list of files, and another to reduce it further, etc... until there are no more `--filter` arguments.

(Also, beware that somebody might apply the same filter twice or thrice, for example:

```console
$ cernopendata-client download-files --recid 5500 --filter-regexp py --filter-regexp level
```

cernopendata_client/cli.py
```diff
@@ -152,7 +158,25 @@ def get_file_locations(server, recid, doi, title, protocol, expand):
     type=click.STRING,
     help="Which CERN Open Data server to query? [default=http://opendata.cern.ch]",
 )
-def download_files(server, recid, doi, title, protocol, expand):
+@click.option(
+    "--filter-name",
+    "name",
+    type=click.STRING,
+    help="Download file that contains the specified string",
```
tiborsimko, line 165: If we look for matching *inside* file name, then the option `--filter-name` plays almost the same role as `--filter-regexp`, and we would be loosing important functionality to match "exactly".  So I suggest to change the meaning of `--filter-name` to make *exact match* for the full filename.  In this way it makes sense to have different options for file name matching. (Although in theory one could do with a single regexp matching option when user would write `^myfilename\.pdf$` but that would be less readable...)

- filter-name ... change to exact matching
- filter-regexp ... good, keep as is
- filter-range -- good, keep as is

When you

cernopendata_client/cli.py
```diff
@@ -152,7 +158,25 @@ def get_file_locations(server, recid, doi, title, protocol, expand):
...
+    "--filter-regexp",
+    "regexp",
+    type=click.STRING,
+    help="Download files matching the regular expression",
+)
+@click.option(
+    "--filter-range",
+    "range",
+    type=click.STRING,
+    help="Download files in a specified range",
```
tiborsimko, line 177: "Download files from a specified list range (i-j)" to make it clearer to users?

cernopendata_client/cli.py
```diff
@@ -161,6 +185,9 @@ def download_files(server, recid, doi, title, protocol, expand):
     \b
     Examples:
       $ cernopendata-client download-files --recid 5500
+      $ cernopendata-client download-files --recid 1 --filter-name 00E081791775.root
```
tiborsimko, line 188: If we change `--filter-name` to do exact match (see above), then please amend example to use full file name.  You can use 5500 which has shorter file names for example.

cernopendata_client/cli.py
```diff
@@ -161,6 +185,9 @@ def download_files(server, recid, doi, title, protocol, expand):
...
+      $ cernopendata-client download-files --recid 1 --filter-regexp E011
```
tiborsimko, line 189: Here too for filter regexp you can use 5500 and '\.py$' perhaps? (to download all Python files could be a good example)

cernopendata_client/downloader.py
```diff
@@ -0,0 +1,73 @@
...
+        print(
+            "==> Downloading file: ./{}/{}".format(
+                path,
+                file_name,
+            )
+        )
+        c = pycurl.Curl()
+        c.setopt(c.URL, file_location)
+        c.setopt(c.WRITEDATA, f)
+        c.setopt(c.NOPROGRESS, False)
```
tiborsimko, line 41: Observation: the progress bar, once a file is downloaded, makes the display broken a bit, see:

```console
$ cernopendata-client download-files --recid 1 --filter-range 1-2
==> Downloading file 1 of 2
==> Downloading file: ./1/0248915F-EE71-E011-8894-0025902009E8.root
==> Downloading file 2 of 2kiB (100%)
==> Downloading file: ./1/0268F635-B671-E011-9090-002481E14E00.root
Downloading: 842237/842237 kiB (100%)
Download completed!
```

Note the `Downloading file 2 of 2kiB (100%)` part where some bits remained from the previous progress bar.

Mentioning just for completeness; if you don't have a quick solution, we can also address this cosmetics issues later via another separate issue... 

However we can solve

tests/test_cli_download_files.py
```diff
@@ -27,3 +27,51 @@ def test_download_files():
...
+
+
+def test_download_files_filter_range():
+    """Test download_files() command with filter-range options."""
+    test_file = "3005/0d0714743f0204ed3c0144941e6ce248.configFile.py"
+    if os.path.isfile(test_file):
+        os.remove(test_file)
+    test_download_files_filter = CliRunner()
+    test_result_regexp = test_download_files_filter.invoke(
+        download_files, ["--recid", 3005, "--filter-range", "0-1"]
```
tiborsimko, line 71: I'm not sure users would have a tendency to start enumerating from 0 as in Python?  Especially if we are printing "Downloading file 1 of 10..."  Perhaps change the logic so that CLI option is passed starting from 1.

tiborsimko: ... also, note that range 0-0 works, 0-1 works, but 1-1 does not do anything:

```console
$ cernopendata-client download-files --recid 3005 --filter-range 1-1

Download completed!
```

ParthS007: It means when user mentions `1`, the file with `0` index will be considered as first file?

tiborsimko: > It means when user mentions `1`, the file with `0` index will be considered as first file?

Yes, exactly.  Basically, when a user asks for `n`-th file, you do `files[n-1]` internally...

cernopendata_client/downloader.py
ParthS007: I tried to reproduce it with other `recid`s but it happens very rarely. I will find a workaround to stop this.

cernopendata_client/cli.py
ParthS007: Now, it's working for multiple filters but not for same filter multiple times. I will create a seperate isssue for it.

tests/test_cli_download_files.py
tiborsimko: The range-starting-from-1 is still not working after latest update.  Here's the use case:

```console
$ cernopendata-client download-files --recid 5500 | head -4
==> Downloading file 1 of 11)
==> Downloading file: ./5500/BuildFile.xml
==> Downloading file 2 of 11
==> Downloading file: ./5500/HiggsDemoAnalyzer.cc
$ cernopendata-client download-files --recid 5500 --filter-range 1
==> Downloading file 1 of 1
==> Downloading file: ./5500/HiggsDemoAnalyzer.cc
Downloading: 81/81 kiB (100%)
Download completed!
```

Note that 2nd file was downloaded, not 1st...

cernopendata_client/cli.py
tiborsimko: The exact filter is not working, here is the use case:

```console
$ cernopendata-client download-files --recid 5500 --filter-name nothing
==> Downloading file 1 of 11
==> Downloading file: ./5500/BuildFile.xml
==> Downloading file 2 of 11
==> Downloading file: ./5500/HiggsDemoAnalyzer.cc
==> Downloading file 3 of 11)
==> Downloading file: ./5500/List_indexfile.txt
==> Downloading file 4 of 11
==> Downloading file: ./5500/M4Lnormdatall.cc
==> Downloading file 5 of 11)
==> Downloading file: ./5500/M4Lnormdatall_lvl3.cc
==> Downloading file 6 of 11)
==> Downloading file: ./5500/demoanalyzer_cfg_level3MC.py
==> Downloading file 7 of 11
==> Downloading file: ./5500/demoanalyzer_cfg_level3data.py
==> Downloading file 8 of 11
==> Downloading file: ./5500/demoanalyzer_cfg_level4MC.py
==> Downloading file 9 of 11
==> Downloading file: ./5500/demoanalyzer_cfg_level4data.py
==> Downloading file 10 of 11
==> Downloading file: ./5500/mass4l_combine.pdf
==> Downloading file 11 of 11
==> Downloading file: ./5500/mass4l_combine.png
Downloading: 90/90 kiB (100%)
Download completed!
``` 

I.e. when no match is found, it downloads everything... it should simply do nothing in such a case.

cernopendata_client/cli.py
```diff
@@ -152,7 +158,25 @@ def get_file_locations(server, recid, doi, title, protocol, expand):
     type=click.STRING,
     help="Which CERN Open Data server to query? [default=http://opendata.cern.ch]",
 )
-def download_files(server, recid, doi, title, protocol, expand):
+@click.option(
+    "--filter-name",
+    "name",
+    type=click.STRING,
+    help="Download file that matching the name",
```
tiborsimko, line 165: "Download files matching exactly the file name"

(there can be several, we'll see it in the future when we shall be having subdirectories...)

cernopendata_client/downloader.py
```diff
@@ -0,0 +1,75 @@
...
+    """Returns the file filtered by the name"""
+    for file_location in file_locations:
+        file_name = file_location.split("/")[-1]
+        if file_name == name:
+            return [file_location]
+
+
+def get_download_files_by_regexp(regexp=None, file_locations=None, **kwargs):
+    """Returns the list of files filtered by a regular expression"""
+    _file_locations = kwargs.get("dload", None)
```
tiborsimko, line 58: What is dload? Would be nice to have some docstring...

cernopendata_client/cli.py
tiborsimko: Ditto for regexp filter:

```console
$ $ cernopendata-client download-files --recid 5500 --filter-regexp Higgs
==> Downloading file 1 of 11
==> Downloading file: ./5500/BuildFile.xml
==> Downloading file 2 of 11
==> Downloading file: ./5500/HiggsDemoAnalyzer.cc
==> Downloading file 3 of 11)
==> Downloading file: ./5500/List_indexfile.txt
==> Downloading file 4 of 11
==> Downloading file: ./5500/M4Lnormdatall.cc
==> Downloading file 5 of 11)
==> Downloading file: ./5500/M4Lnormdatall_lvl3.cc
==> Downloading file 6 of 11)
==> Downloading file: ./5500/demoanalyzer_cfg_level3MC.py
==> Downloading file 7 of 11
==> Downloading file: ./5500/demoanalyzer_cfg_level3data.py
==> Downloading file 8 of 11
==> Downloading file: ./5500/demoanalyzer_cfg_level4MC.py
==> Downloading file 9 of 11
==> Downloading file: ./5500/demoanalyzer_cfg_level4data.py
==> Downloading file 10 of 11
==> Downloading file: ./5500/mass4l_combine.pdf
==> Downloading file 11 of 11
==> Downloading file: ./5500/mass4l_combine.png
Downloading: 90/90 kiB (100%)
Download completed!
```

Should download only some files...

cernopendata_client/downloader.py
ParthS007: Added a doc string.

cernopendata_client/cli.py
ParthS007: Added this edge case when there are filters applied and no files match with the filters, no files will be downloaded

tiborsimko (approved):

PR #50, opened 2020-09-17
tests: download-files filters using recid 5500

### LINKED ISSUE #47: tests: download-files filters could use record 5500
(1) The test suite for the filtering functionality (see `test_download_files_filter_*()`) currently use the test record 3005. However, this record contains only a single file, which cannot test various scenarios (such as getting files 1-3).

It would be therefore good to amend these tests to use record 5500 which has several inputs files to choose from.  It could also  test exactly the same examples that we promise in the docstring when somebody does `cernopendata-client download-files --help`.

That said, the filters work fine :+1: this issue is only about enriching tests to be able to have more files to filter from to cover more usage scenarios. 

(2) One could also add three more simple test such as filtering for a non-existing file name (`--filter-name notexisting`) should return error message, and (2b) the same for regexp and (2c) the same for fwrong range (`--filter-range 0` and `--filter-range 99998-99999`).  Albeit the last one is kind of tested elsewhere in `test_validate_range()` as well.

ParthS007:
closes #47

### DIFF
--- a/cernopendata_client/validator.py
+++ b/cernopendata_client/validator.py
@@ -41,6 +41,11 @@ def validate_range(range=None, count=None):
     Return True if range lies in total files count,
     raises click.BadParameter exception otherwise.
     """
+    if len(range.split("-")) != 2:
+        raise click.BadParameter(
+            "Range should have start and end index(i-j)",
+            param_hint=["--filter-range"],
+        )
     range_from, range_to = int(range.split("-")[0]), int(range.split("-")[-1])
     if range_from <= 0:
         raise click.BadParameter(
--- a/tests/test_cli_download_files.py
+++ b/tests/test_cli_download_files.py
@@ -10,6 +10,7 @@
 """cernopendata-client cli command download-files test."""
 
 import os
+import pytest
 from click.testing import CliRunner
 from cernopendata_client.cli import download_files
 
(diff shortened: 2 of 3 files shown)


tests/test_cli_download_files.py
```diff
@@ -57,27 +63,40 @@ def test_download_files_filter_regexp():
     if os.path.isfile(test_file):
         os.remove(test_file)
     test_download_files_filter = CliRunner()
-    test_result_regexp = test_download_files_filter.invoke(
+    test_result_regexp_one = test_download_files_filter.invoke(
         download_files, ["--recid", 3005, "--filter-regexp", "py$"]
```
tiborsimko, line 67: A good test could be also to get `.cc` files from record 5500, because they are three, and they are not sequential, i.e. this would test whether we can pick up files from random list locations :wink:

tests/test_cli_download_files.py
```diff
@@ -57,27 +63,40 @@ def test_download_files_filter_regexp():
...
     )
-    assert test_result_regexp.exit_code == 0
+    assert test_result_regexp_one.exit_code == 0
     assert os.path.isfile(test_file) is True
-    assert test_result_regexp.output.endswith("\nDownload completed!\n")
+    assert test_result_regexp_one.output.endswith("\nDownload completed!\n")
     if os.path.isfile(test_file):
         os.remove(test_file)
+    test_result_regexp_two = test_download_files_filter.invoke(
+        download_files, ["--recid", 5500, "--filter-regexp", "0"]
```
tiborsimko, line 75: To test non-existent regexp for files, please use longer string such as "notexisting" or "wontmatchanything" or something... Just in case there could be an update to record 5500 that would bring a file containing 0 inside.

cernopendata_client/validator.py
```diff
@@ -41,6 +41,11 @@ def validate_range(range=None, count=None):
     Return True if range lies in total files count,
     raises click.BadParameter exception otherwise.
     """
+    if len(range.split("-")) != 2:
```
tiborsimko, line 44: Hmm, currently with `master` version the single digit input also works fine; I mean that this:

```
 $ cernopendata-client download-files --recid 5500 --filter-range 1
```

is the same as this:

```
 $ cernopendata-client download-files --recid 5500 --filter-range 1-1
```

Which is perhaps nice to keep?

However, we can also require to always have A-B as you mention, so that the notion of "range" is ultra clear :+1:

ParthS007: I think I will prefer keeping the start and end in the `range` option else users can always use the `name` option if they want to download a specific file.

But I am open to adding only one index in range.

tests/test_cli_download_files.py
```diff
@@ -31,53 +32,92 @@ def test_download_files():
...
+    if os.path.isfile(test_one_file):
+        os.remove(test_one_file)
+    else:
+        pytest.fail("{} not downloaded".format(test_one_file))
+
+    for file_location in test_two_files:
+        if os.path.isfile(file_location):
+            os.remove(file_location)
+    test_result_regexp_two = test_download_files_filter.invoke(
+        download_files, ["--recid", 5500, "--filter-regexp", "\.cc$"]
```
tiborsimko, line 85: I get one pytest warning on this:

```
tests/test_cli_download_files.py:85
  /home/simko/private/project/opendata/src/cernopendata-client/tests/test_cli_download_files.py:85: DeprecationWarning: invalid escape sequence \.
    download_files, ["--recid", 5500, "--filter-regexp", "\.cc$"]

...

16 passed, 1 warning in 7.67s
```

Can you please fix? Would be nice if all tests are reported as passing.

ParthS007: It is because of `\` in the regexp!

PR #51, opened 2020-09-18
cli: get-metadata: expose only metadata

### LINKED ISSUE #8: get-metadata: expose only metadata
Due to special treatment of files, some of the default API output (notably file buckets) are not useful, since the files are being served from EOS.

These should be hidden.  We should probably return directly the content of `metadata`.

ParthS007:
closes #8

### DIFF
--- a/cernopendata_client/cli.py
+++ b/cernopendata_client/cli.py
@@ -96,7 +96,14 @@ def get_metadata(server, recid, doi, title, output_fields):
             )
             sys.exit(1)
     else:
-        output_json = record_json
+        output_json = record_json["metadata"]
+        if "_files" in output_json:
+            del output_json["_files"]
+        for field in output_json["files"]:
+            if "bucket" in field:
+                del field["bucket"]
+            if "version_id" in field:
+                del field["version_id"]
     click.echo(json.dumps(output_json, indent=4))
 
 
--- a/tests/test_cli_get_metadata.py
+++ b/tests/test_cli_get_metadata.py
@@ -22,6 +22,10 @@ def test_get_metadata_from_recid():
         '"title": "Configuration file for LHE step HIG-Summer11pLHE-00114_1_cfg.py"'
         in test_result.output
     )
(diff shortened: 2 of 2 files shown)


cernopendata_client/cli.py
```diff
@@ -96,7 +96,7 @@ def get_metadata(server, recid, doi, title, output_fields):
             )
             sys.exit(1)
     else:
-        output_json = record_json
+        output_json = record_json["metadata"]
```
tiborsimko, line 99: Works fine :+1: We could also delete `_files.bucket` and `_files.version_id` because people cannot use them...

```
  "_files": [
        {
            "bucket": "61319316-0704-4b6d-be7c-14017099015c",
            "checksum": "adler32:ff63668a",
            "key": "BuildFile.xml",
            "size": 305,
            "version_id": "713b7316-a48a-4a20-981d-7836740f4f8b"
        },
```

cernopendata_client/cli.py
```diff
@@ -96,7 +96,12 @@ def get_metadata(server, recid, doi, title, output_fields):
...
+        for field in output_json["_files"]:
+            if "bucket" in field:
+                del field["bucket"]
+            if "version_id" in field:
+                del field["version_id"]
```
tiborsimko, line 104: Works well :+1: Can you please add simple assert into existing tests to say that these fields ("bucket" etc) should not be present in the output?

ParthS007: okay 👍

ParthS007: @tiborsimko There are two similar fields namely `_files` and `files` which have the same data. Should I remove these 2 keys from `files` as well?

tiborsimko: Indeed, good catch! You can therefore remove the whole `_files` field, let's keep only `files` and let's remove `bucket` and `version_id` from there. 

P.S. Writing tests has a nice side effect in spotting these things :wink:

ParthS007: Yeah, I agree 😃

tiborsimko (approved):

PR #52, opened 2020-09-18
cli: get-metadata - amend `--output-fields` behaviour

### LINKED ISSUE #41: get-metadata: amend `--output-fields` behaviour
Stemming from #8 , we should also fix the behaviour of `--output-fields`.  This currently allows to select only `metadata`,  `created`, ... and other top-level fields from the `/api/records` output.  This is not very useful.

While the power users can do things like:

```console
$ cernopendata-client get-metadata --recid 1 | jq -S '.metadata.title'
"/BTau/Run2010B-Apr21ReReco-v1/AOD"
$ cernopendata-client get-metadata --recid 1 | jq -S '.metadata.system_details.global_tag'
"FT_R_42_V10A::All"
```

it would be useful to amend `--output-fields` to allow searching inside metadata section, so that users could write instead:

```console
$ cernopendata-client get-metadata --recid 1 --output-fields title
/BTau/Run2010B-Apr21ReReco-v1/AOD
$ cernopendata-client get-metadata --recid 1 --output-fields system_details.global_tag
FT_R_42_V10A::All
```

IOW, `--output-fields` should be able to print out a wanted JSON path value directly.

ParthS007:
closes #41

- Added searching of fields in nested json
- Added tests and docs

### DIFF
--- a/cernopendata_client/cli.py
+++ b/cernopendata_client/cli.py
@@ -52,11 +52,10 @@ def version():
 @click.option("--doi", help="Digital Object Identifier")
 @click.option("--title", help="Title of the record")
 @click.option(
-    "--output-fields",
+    "--output-value",
     is_flag=False,
     type=click.STRING,
-    help="Comma-separated list of fields from the record "
-    "that should be included in the output.",
+    help="Value of fields from the record that should be included in the output.",
 )
 @click.option(
     "--server",
@@ -64,7 +63,7 @@ def version():
     type=click.STRING,
     help="Which CERN Open Data server to query? [default=http://opendata.cern.ch]",
 )
-def get_metadata(server, recid, doi, title, output_fields):
+def get_metadata(server, recid, doi, title, output_value):
     """Get metadata content of a record.
 
     Select a CERN Open Data bibliographic record by a record ID, a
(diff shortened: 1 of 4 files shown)


cernopendata_client/cli.py
```diff
@@ -73,37 +73,33 @@ def get_metadata(server, recid, doi, title, output_fields):
     \b
     Examples:
       $ cernopendata-client get-metadata --recid 1
-      $ cernopendata-client get-metadata --recid 1 | jq -S '.metadata.title'
+      $ cernopendata-client get-metadata --recid 1 system_details.global_tag
```
tiborsimko, line 76: The example would need `--output-fields` leading flag.

cernopendata_client/cli.py
```diff
@@ -73,37 +73,33 @@ def get_metadata(server, recid, doi, title, output_fields):
...
     record_json = get_record_as_json(server, recid, doi, title)
+    output_json = record_json["metadata"]
     if output_fields is not None:
-        output_fields = [f.strip() for f in output_fields.split(",")]
-    if output_fields and len(output_fields) > 0:
-        try:
-            output_json = {field: record_json[field] for field in output_fields}
-        except KeyError:
-            top_level_fields = ", ".join(field for field in record_json)
+        fields = output_fields.split(".")
```
tiborsimko, line 84: The docs promises the output can be a list of wanted things, so note that `output_fields` can be a list itself.  The PR does currently this:

```console
$ cernopendata-client get-metadata --recid 5500 --output-fields recid,title
Key 'recid,title' is not present in metadata
```

One would expect to have listed both fields in the output.

docs/usage.rst
```diff
@@ -60,15 +60,18 @@ contain interesting physics information describing the dataset.
...
-    "/BTau/Run2010B-Apr21ReReco-v1/AOD"
-    $ cernopendata-client get-metadata --recid 1 | jq -S '.metadata.system_details.global_tag'
-    "FT_R_42_V10A::All"
+    $ cernopendata-client get-metadata --recid 1 --output-fields title
+    {
+      "title" : "/BTau/Run2010B-Apr21ReReco-v1/AOD"
+    }
+    $ cernopendata-client get-metadata --recid 1 --output-fields system_details.global_tag
+    {
+      "global_tag": "FT_R_42_V10A::All"
```
tiborsimko, line 73: Users would want to get only the value, so that they can use `cernopendata-client` is shell scripts like:

```
global_tag=$(cernopendata-client get-metadata --recid $myrecid --output-fields 'system_details.global_tag')
```

If we always output JSON, then people would still need to use something like `jq` to parse the output.

Two options:

- either add an option to get values only (such as `--output-values`)
- or always output only values and do not allow to specify multiple fields (so that output is always just text); and let people who want to do more complex stuff could always use `jq`.

I guess the latter seems nicer, but we would have to rename the option `--output-values`.

ParthS007: After discussing IRL, renaming to `--output-values` and getting value only. 👍

tiborsimko: (1) Actually we are getting the JSON value i.e. also with the string quotes... Perhaps we should get rid of those...

```console
$ cernopendata-client get-metadata --recid 1 --output-values system_details.global_tag
"FT_R_42_V10A::All"
```

I guess users would expect to see just the value, as mentioned in the #41 issue decription:

We could do:

- if output is number: nothing to do, print as is
- if string: strip quotes
- if list of dict: print as is

(2) Also, since we don't allow multiple values, please do:

- rename option from `--output-values` to `--output-value`
- amend help strings which currently still say "Comma-separated list of fields...", see `cernopendata-client get-metadata --help`

tiborsimko (approved):

PR #54, opened 2020-09-23
cli: initial implementation of verify-files

### LINKED ISSUE #23: verify-files: initial release
After user downloads a set of files belonging to a dataset, see #22, it will be interesting to provide a functionality to check the sizes and adler32 checksums of downloaded files.

This information is available over REST API, here is [one example](http://opendata.cern.ch/record/5500/export/json). 

The user would launch:

```console
$ cernopendata-client verify-files --recid 5500 --directory ./mydata/5550
```

The command would go through the given directory, calculate size and adler32-checksum of present files, and compare this with the information obtained from the server via the REST API.

The command would report back to user:
- exit code 1 if any of the locally available files is having different size or different checksum
- exit code 2 if some of the locally available files is not present on the remote opendata.cern.ch instance
- exit code 3 if some of the opendata.cern.ch files are not present locally
- exit code 0 if everything corresponds perfectly well

The command will exit with different error codes so that users could plug this command to their harvesting workflows.

tiborsimko:
Adds command ``verify-files`` with initial documentation and test
suite. Closes #23.

Signed-off-by: Tibor Šimko <tibor.simko@cern.ch>


cernopendata_client/cli.py
```diff
@@ -257,3 +257,79 @@ def download_files(server, recid, doi, title, protocol, expand, name, regexp, ra
...
+    verify integrity of downloaded data files belonging to this
+    record.
+
+    \b
+    Examples:
+      $ cernopendata-client verify-files --recid 5500
+    """
+
+    from .verifier import get_file_info_local
+    from .search import get_file_info_remote
```
ParthS007, line 283: Just curious, why this is in-function import?

cernopendata_client/search.py
```diff
@@ -136,3 +136,29 @@ def get_files_list(server=None, record_json=None, protocol=None, expand=None):
...
+def get_file_info_remote(recid):
+    """Return remote file information list for given record.
+
+    Returns a list of dictionaries containing (checksum, name, size,
+    uri) for each file in the record.  Note that 'name' property is
+    not stored remotely, but is calculated in this function for
+    convenience.
+    """
+    file_info_remote = []
+    record_json = get_record_as_json(server="http://opendata.cern.ch", recid=recid)
```
ParthS007, line 150: Why `server` is hardcoded here?

ParthS007 (approved):
Works like wonder 🚀 . Just a couple of questions I was curious about. :)

cernopendata_client/search.py
tiborsimko: Need to propagate `server` variable; thought of doing that while refactoring.

cernopendata_client/cli.py
tiborsimko: Yes, I can move them "upstairs".

cernopendata_client/search.py
ParthS007: okay, makes sense 👍

ParthS007:
Screenshot 2020-09-24 at 16 57 49
I see this warning when I run tests locally, is it relevant?

ParthS007 (approved):
LGTM!

tiborsimko:
> I see this warning when I run tests locally, is it relevant? 

I don't see it locally. Note the `pep8.py` file... What about when you run from a clean slate?

```
$ git clean -d -ff -x
$ ./run-tests.sh
```

PR #68, opened 2020-10-13
ci: add shellcheck and improve run-tests script

### LINKED ISSUE #64: Improve the run-tests shell script
Later improvement: instead of having commands written twice, we could define shell functions like:

```shell

check_black () {
    black --check .
}
```

so that we can simply call `check_black` in appropriate places here...

In this way the commands to run will be defined in a single place (inside the shell function).

_Originally posted by @tiborsimko in https://github.com/cernopendata/cernopendata-client/pull/36#discussion_r503300017_

### LINKED ISSUE #66: ci: shellcheck
It would be good to add `shellcheck` to test `run-tests.sh` itself:

- add it as a target `run-tests.sh --check-shell`
- add it to GA as a check

ParthS007:
closes #64
closes #66

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,6 +7,16 @@ on:
         branches: master
 
 jobs:
+    Lint-Shellcheck:
+        runs-on: ubuntu-latest
+        steps:
+          - uses: actions/checkout@v2
+
+          - name: Runs shell script static analysis
+            run: |
+              sudo apt-get install shellcheck
+              ./run-tests.sh --check-shellscript
+
     Lint-black:
         runs-on: ubuntu-latest
         steps:
--- a/run-tests.sh
+++ b/run-tests.sh
@@ -13,37 +13,54 @@ set -o errexit
 # Quit on unbound symbols
 set -o nounset
 
+check_script () {
+    shellcheck run-tests.sh
+}
+
+check_black () {
+    black --check .
+}
+
+check_pydocstyle () {
+    pydocstyle cernopendata_client
+}
+
+check_flake8 () {
+    flake8 .
+}
+
+check_manifest () {
+    check-manifest
+}
+
+check_sphinx () {
+    sphinx-build -qnNW docs docs/_build/html
+    sphinx-build -qnNW -b doctest docs docs/_build/doctest
+}
+
+check_pytest () {
+    python setup.py test
+}
+
 for arg in "$@"
 do
     case $arg in
-        --check-black) # Check Python code formatting
-        black --check .
-        ;;
-        --check-pydocstyle) # Check compliance with Python docstring conventions
-        pydocstyle cernopendata_client
-        ;;
-        --check-flake8) # Check compliance with pep8, pyflakes and circular complexity
-        flake8 .
-        ;;
-        --check-manifest) # Check Python manifest completeness
-        check-manifest
-        ;;
-        --check-sphinx) # Check Sphinx documentation with doctests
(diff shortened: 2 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -7,6 +7,16 @@ on:
         branches: master
 
 jobs:
+    Shell-check:
```
tiborsimko, line 10: `shellcheck` is another linter, like flake8... So let's use `Lint-shellcheck` as the visible name.

run-tests.sh
```diff
@@ -13,36 +13,67 @@ set -o errexit
 # Quit on unbound symbols
 set -o nounset
 
+check_script () {
+    shellcheck run-tests.sh
+}
```
tiborsimko, line 18: Does the GA output looks nicer if we add those explicit echo commands? For example:

```
check_script () {
    echo "==> Checking scripts..."
    shellcheck run-tests.sh
    echo "==> Success!"
}
```

Maybe we don't need that... I thinking out loud how we could improve the output that currently in case of success looks somewhat cryptic:

2020-10-13-143025_2560x1440_scrot

run-tests.sh
```diff
@@ -13,36 +13,67 @@ set -o errexit
...
+
+check_pytest () {
+    python setup.py test
+}
+
 for arg in "$@"
 do
     case $arg in
+        --check-shellscript) # Runs shell script static analysis
```
tiborsimko, line 48: These comments can go into functions (or, if transformed as `echo` statement, can even go away) so that here the `case` program can be nicely simplified... e.g. all in one line...

.github/workflows/ci.yml
ParthS007: Done 👍

run-tests.sh
ParthS007: @tiborsimko I have these comments as GA steps name in CI so I can remove the comments from shell script.

run-tests.sh
```diff
@@ -13,36 +13,53 @@ set -o errexit
...
-        sphinx-build -qnNW -b doctest docs docs/_build/doctest
-        python setup.py test
+        --check-shellscript) check_script;;
+        --check-black) check_black;;
+        --check-pydocstyle) check_pydocstyle;;
+        --check-flake8) check_flake8;;
+        --check-manifest) check_manifest;;
+        --check-sphinx) check_sphinx;;
+        --pytest) check_pytest;;
+        --all)
```
tiborsimko, line 55: Perhaps instead of `--all`, we should use `--check-all`, for better harmony with the rest...

tiborsimko: Can you pleaso also indent the commands under target in the usual shell script way?

```
for arg in "$@"
do
    case $arg in
        --check-shellscript) check_script;;
        --check-black) check_black;;
        --check-pydocstyle) check_pydocstyle;;
        --check-flake8) check_flake8;;
        --check-manifest) check_manifest;;
        --check-sphinx) check_sphinx;;
        --pytest) check_pytest;;
        --all)
            check_script
            check_black
            check_pydocstyle
            check_flake8
            check_manifest
            check_sphinx
            check_pytest
            ;;
        *)
    esac
done
```

run-tests.sh
ParthS007: Added a black error.

tiborsimko: Looks good; please remove the black error and we are good to go I think

ParthS007: Done 👍

tiborsimko (approved):

PR #70, opened 2020-10-16
cli: printer initial release

### LINKED ISSUE #61: printer: initial release
Instead of having `print()` and `click.echo()` hard-coded throughout the code base, let's introduce a new `printer.py` module that can take care of displaying info/error messages in unified manner.

Example: `display_message()` function in `reana-dev`.

We could then print success in green, progress in yellow, errors in red, etc.

The code base should then be amended to use the new `display_message()` function instead of printing or echoing.

ParthS007:
- replace `print()` and `click.echo()` with `display_message()`

closes #61

### DIFF
--- a/cernopendata_client/printer.py
+++ b/cernopendata_client/printer.py
@@ -1,0 +1,52 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of REANA.
+# Copyright (C) 2020 CERN.
+#
+# cernopendata-client is free software; you can redistribute it and/or modify
+# it under the terms of the GPLv3 license; see LICENSE file for more details.
+
+"""cernopendata-client output print configuration."""
+
+import click
+
+from .config import PRINTER_COLOUR_ERROR, PRINTER_COLOUR_INFO, PRINTER_COLOUR_SUCCESS
+
+
+def display_message(prefix=None, msg_type=None, msg=None):
+    """Display message in a similar style as run_command().
+
+    :param prefix: prefix to be added in output (==> / ->)
+    :param msg_type: Type of message - Error/Progress/Success
+    :param msg: message to display
+    :type prefix: str
+    :type msg_type: str
+    :type msg: str
+    """
+    msg_color_map = {
+        "info": PRINTER_COLOUR_INFO,
+        "error": PRINTER_COLOUR_ERROR,
+        "success": PRINTER_COLOUR_SUCCESS,
+    }
+    msg_color = msg_color_map.get(msg_type, None)
+
+    if prefix == "double":
+        if msg_color:
+            click.secho("==> ", bold=True, nl=False, fg="{}".format(msg_color))
+        else:
+            click.secho("==> ", bold=True, nl=False)
+
+        if msg_type == "error":
+            click.secho(
+                "{}: ".format(msg_type.upper()),
+                bold=True,
+                nl=False,
+                fg=PRINTER_COLOUR_ERROR,
+            )
+    elif prefix == "single":
+        click.secho("  -> ", bold=False, nl=False)
+
+    if msg_color:
+        click.secho("{}".format(msg), bold=True, nl=True, fg="{}".format(msg_color))
+    else:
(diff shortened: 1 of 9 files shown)


cernopendata_client/printer.py
```diff
@@ -0,0 +1,44 @@
...
+    :type prefix: str
+    :type msg_type: str
+    :type msg: str
+    """
+    msg_color_map = {
+        "success": "bright_green",
+        "progress": "bright_yellow",
+        "error": "red",
+    }
+    if prefix:
```
tiborsimko, line 29: My principal comment is that the leading prefix `==>` is printed in different colour than the rest of the headline, for example:

2020-10-19-095745_586x478_scrot

- in case of errors, the prefix `==>` should be red, to better visually indicate the nature of the report
- in case of info messages, ditto, the prefix `==>` should be in yellow
- the secondary prefix `->` is less important, so it can always stay in green; only the leading prefix should be coloured depending on the info vs error content

tiborsimko: (also, to use leading green colour for `==>` is nice, but the leading headline should then continue, e.g. simply in regular colour but in **bold**, or else continue with the green)

cernopendata_client/printer.py
```diff
@@ -0,0 +1,44 @@
...
+    """Display message in a similar style as run_command().
+
+    :param prefix: prefix to be added in output (==> / ->)
+    :param msg_type: Type of message - Error/Progress/Success
+    :param msg: message to display
+    :type prefix: str
+    :type msg_type: str
+    :type msg: str
+    """
+    msg_color_map = {
```
tiborsimko, line 24: This could be moved to the global `config.py` now that we have it, e.g.

```python
PRINTER_COLOUR_INFO='green'
PRINTER_COLOUR_ERROR='red'
```

cernopendata_client/printer.py
tiborsimko: BTW the same for success:

cernopendata_client/printer.py
```diff
@@ -0,0 +1,44 @@
...
+#
+# cernopendata-client is free software; you can redistribute it and/or modify
+# it under the terms of the GPLv3 license; see LICENSE file for more details.
+
+"""cernopendata-client output print configuration."""
+
+import click
+
+
+def display_message(prefix=None, msg_type=None, msg=None):
```
tiborsimko, line 14: Please also hunt for the other occurrences of printing so that we centralise `==>` for all messages, e.g. we currently  have:

```console
$ cernopendata-client download-files

ERROR: Please provide at least one of following arguments: (recid, doi, title)
```

which is not consistent with:

```console
$ cernopendata-client verify-files --recid 5500 --server http://opendata.cern.ch
...
==> Verifying file demoanalyzer_cfg_level4MC.py...
  -> expected size 3874, found 3864
==> ERROR: File size does not match.
```

I would expect to see `==> ERROR` everywhere.

cernopendata_client/printer.py
ParthS007: Done 👍

ParthS007: Done 👍

cernopendata_client/printer.py
ParthS007: Done 👍

cernopendata_client/printer.py
ParthS007: Done 👍

tiborsimko: Actually I see this:

2020-10-19-140222_582x130_scrot

where the final Success is printed in the same colour as regular text... so it is less graphically apparent than the other Downloading/Verifying headlines, while it should be at least equally well represented.

Note also that the "-> " in Progress is not printed in bold as "-> " in Files.

ParthS007: Now, `Success!` is equally apparent to other headings.

`->` is now same everywhere (not bold).

ParthS007: Done 👍

tiborsimko (approved):

PR #75, opened 2020-10-26
cli: download-files - xrootd protocol

### LINKED ISSUE #24: download-files: xrootd protocol
In addition to http protocol implemented in #22, we should add support for xrootd protocol which provides more bandwidth.

Check whether client system contains xrootd commands (`xrdcp`) or Python libraries (`xrootd`).

If yes, enable support for `--protocol root`.

If not, report help string what people has to do do install them.

This is not an MVP functionality for the first public release; it can be added later.

ParthS007:
closes #24

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,32 +5,49 @@
 # cernopendata-client is free software; you can redistribute it and/or modify
 # it under the terms of the GPLv3 license; see LICENSE file for more details.
 
-# Use Python 3.8 slim base image
-FROM python:3.8-slim
+# Use CentOS8
+FROM centos:8
 
 # Install system prerequisites
-RUN apt-get update -y && \
-    apt-get install -y \
+RUN yum update -y && \
+    yum install -y \
+        ca-certificates \
+        epel-release && \
+    yum groupinstall -y "Development Tools" && \
+    yum install -y \
+        cmake \
         curl \
         gcc \
-        libcurl4-openssl-dev \
-        libssl-dev && \
-    apt-get autoremove && \
-    apt-get clean
+        gcc-c++ \
+        python3 \
+        python3-pip \
+        python3-devel \
+        libcurl-devel \
+        zlib-devel \
+        libuuid-devel \
+        openssl-devel && \
+    yum autoremove && \
+    yum clean all
+
+# Install some prerequisites ahead of `setup.py` in order to take advantage of
+# the docker build cache:
+RUN pip3 install --upgrade pip setuptools
+RUN pip3 install wheel
 
 # Add sources to `/code` and work there
 WORKDIR /code
 ADD . /code
 
 # Install cernopendata-client
-RUN pip install .
+RUN pip3 install .[xroot]
 
 # Remove /code to make image slimmer
 RUN rm -rf /code
 
 # Run container as `cernopendata` user with UID `1000`, which should match
 # current host user in most situations
-RUN adduser --uid 1000 --disabled-password --gecos '' cernopendata
+RUN adduser --uid 1000  cernopendata --gid 0 && \
+    chown -R  cernopendata:root /code
 
 # Run cernopendata-client upon entry
 USER cernopendata
--- a/cernopendata_client/downloader.py
+++ b/cernopendata_client/downloader.py
@@ -8,9 +8,17 @@
 
 """cernopendata-client file downloading related utilities."""
 
+import sys
 from sys import stderr as STREAM
 import re
 import pycurl
+
+try:
+    from xrootdpyfs import XRootDPyFS
+
+    xrootd_available = True
+except ImportError:
+    xrootd_available = False
 
 from .validator import validate_range
 from .printer import display_message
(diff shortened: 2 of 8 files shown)


tests/test_cli_download_files.py
```diff
@@ -68,6 +68,23 @@ def test_download_files():
         os.remove(test_file)
 
 
+def test_download_files_root():
```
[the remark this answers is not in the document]
tiborsimko: (Well, you could actually also amend test behaviour so that it would not result in E if xrootd is not available, and only test root downloads if xrootd is present.)

Dockerfile
```diff
@@ -5,32 +5,53 @@
...
+        python3 \
+        python3-pip \
+        python3-devel \
+        libcurl-devel \
+        zlib-devel \
+        libuuid-devel \
+        openssl-devel && \
+    yum autoremove && \
+    yum clean all
+
```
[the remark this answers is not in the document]
ParthS007: Done 👍

Dockerfile
```diff
@@ -5,32 +5,53 @@
...
 # Add sources to `/code` and work there
 WORKDIR /code
 ADD . /code
 
 # Install cernopendata-client
-RUN pip install .
+RUN pip3 install .
+
+# Install xrootdpyfs from PyPI
+RUN pip3 install .[xroot]
```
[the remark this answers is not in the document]
ParthS007: Done 👍

cernopendata_client/downloader.py
```diff
@@ -8,10 +8,18 @@
...
 
+import sys
 from sys import stderr as STREAM
 import re
 import pycurl
 
+try:
+    from xrootdpyfs import XRootDPyFS
+
+    xrootd_warning = False
```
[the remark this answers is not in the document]
ParthS007: Done 👍

cernopendata_client/downloader.py
```diff
@@ -29,26 +37,49 @@ def show_download_progress(download_t, download_d, upload_t, upload_d):
...
+            )
+            c = pycurl.Curl()
+            c.setopt(c.URL, file_location)
+            c.setopt(c.WRITEDATA, f)
+            c.setopt(c.NOPROGRESS, False)
+            c.setopt(c.XFERINFOFUNCTION, show_download_progress)
+            c.perform()
+            c.close()
+        print()
+    elif protocol == "root":
```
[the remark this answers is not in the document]
ParthS007: Done 👍

tests/test_cli_download_files.py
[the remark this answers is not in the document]
ParthS007: Done 👍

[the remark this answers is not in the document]
ParthS007: I tried to implement the test, pushed to check if it works on `gh-actions`!

setup.py
```diff
@@ -24,6 +24,7 @@
     "pytest-cache>=1.0",
     "pytest-cov>=1.8.0",
     "pytest>=2.8.0",
+    "mock==3.0.5",
```
tiborsimko, line 27: We should be also able to say here:

```
"mock>=3.0,<4.0",
```

which will open a possibility for some bug fix releases perhaps.

tests/test_cli_download_files.py
```diff
@@ -68,6 +68,37 @@ def test_download_files():
...
+    import mock
+    import sys
+
+    with mock.patch.dict(sys.modules, {"xrootdpyfs": None}):
+        test_download_files = CliRunner()
+        test_result = test_download_files.invoke(
+            download_files, ["--recid", 3005, "--protocol", "root"]
+        )
+        assert test_result.exit_code == 1
+        assert "xrootd is not installed on system" in test_result.output
```
tiborsimko, line 99: This test fails for me (and I do have xrootd locally available):

```
>           assert test_result.exit_code == 1
E           assert 0 == 1
E            +  where 0 = <Result okay>.exit_code

tests/test_cli_download_files.py:98: AssertionError
```

This is similar to the failure on Travis CI... So the mocking is not fully working?

Alternatively, you can import `xrootd_available` variable, and check based on that...

ParthS007: Yes, The mocking is not working correctly, since when I saw examples over the web mocking of packages in `sys. modules` is possible but `xrootdpyfs` is not installed in `sys.modules` directly but in `python-packages`. I am looking for a workaround for it.

cernopendata_client/downloader.py
```diff
@@ -29,26 +37,49 @@ def show_download_progress(download_t, download_d, upload_t, upload_d):
...
+        if not xrootd_available:
+            display_message(
+                prefix="double",
+                msg_type="error",
+                msg="xrootd is not installed on system.",
```
tiborsimko, line 66: You can also add to the error message "Please use the http protocol instead." to inform the users about what they should do instead.

```
==> ERROR: XRootD is not installed on system. Please use the 'http' protocol instead.
```

setup.py
ParthS007: Done 👍

cernopendata_client/downloader.py
ParthS007: Done 👍

tests/test_cli_download_files.py
ParthS007: @tiborsimko Now the testing of this case is working fine 🎉 . I tested with multiple ways and it seems to work out perfect.

tiborsimko: There is one way where it fails locally for me:

```console
$ mkvirtualenv test
$ pip install . 
$ ./run-tests.sh --check-pytest

...

    def test_download_files_root():
        """Test download_files() command with root protocol."""
        test_file = "3005/0d0714743f0204ed3c0144941e6ce248.configFile.py"
        if os.path.isfile(test_file):
            os.remove(test_file)
        test_download_files = CliRunner()
        test_result = test_download_files.invoke(
            download_files, ["--recid", 3005, "--protocol", "root"]
        )
>       assert test_result.exit_code == 0
E       assert 1 == 0
E        +  where 1 = <Result SystemExit(1)>.exit_code

tests/test_cli_download_files.py:81: AssertionError

===== 1 failed, 52 passed in 14.95s =====
```

That is, when I install with xrootd, everything works and 53 tests pass.  But when I install without xrootd, one test fails.  I would have expected the test to say: "xrootd not available, so I'm not running".  Basically you can try to import  `xrootd_available` and [skip this test](https://docs.pytest.org/en/stable/skipping.html#skipping-test-functions) when it is not applicable on the given platform.

tiborsimko: One more thing: Can you please enrich `docs/usage.rst` with a tiny `download-files --protocol root` example?

ParthS007: @tiborsimko I think it will be more apt to use this [importskip](https://docs.pytest.org/en/stable/skipping.html#skipping-on-a-missing-import-dependency) here.

tiborsimko (approved):

PR #77, opened 2020-11-03
docs: enrich CLI API page

### LINKED ISSUE #55: docs: enrich CLI API page
Let's enrich the CLI API docs page with all commands and options that would be automatically generated.

```
I thought of looking into generating the CLI API in other way... E.g. we could run click-sphinx
locally and add the file into `docs` so that ReadTheDocs would simply consume it from 
there.  (A bit like we are doing with OpenAPI stuff in REANA.) Hence I used `addresses` 
and not `closes` yet... I'll move the issue into "Ready for work" kanban column.
```

_Originally posted by @tiborsimko in https://github.com/cernopendata/cernopendata-client/issues/40#issuecomment-694109978_

ParthS007:
closes #55

### DIFF
--- a/cernopendata_client/cli.py
+++ b/cernopendata_client/cli.py
@@ -80,7 +80,7 @@ def get_metadata(server, recid, doi, title, output_value):
     \b
     Examples:
       $ cernopendata-client get-metadata --recid 1
-      $ cernopendata-client get-metadata --recid 1 --output-value system_details.global_tag
+      $ cernopendata-client get-metadata --recid 1 --output-value title
     """
     validate_server(server)
     if recid is not None:
@@ -211,12 +211,10 @@ def download_files(
     Examples:
       $ cernopendata-client download-files --recid 5500
       $ cernopendata-client download-files --recid 5500 --filter-name BuildFile.xml
-      $ cernopendata-client download-files --recid 5500 --filter-name BuildFile.xml,List_indexfile.txt
       $ cernopendata-client download-files --recid 5500 --filter-regexp py$
       $ cernopendata-client download-files --recid 5500 --filter-range 1-4
       $ cernopendata-client download-files --recid 5500 --filter-range 1-2,5-7
       $ cernopendata-client download-files --recid 5500 --filter-regexp py --filter-range 1-2
-      $ cernopendata-client download-files --recid 5500 --filter-regexp py --filter-range 1-2,3-4
     """
     validate_server(server)
     if recid is not None:
--- a/docs/cliapi.rst
+++ b/docs/cliapi.rst
@@ -1,11 +1,29 @@
-.. _cliapi:
-
 CLI API
 =======
 
-Please use ``cernopendata-client --help`` in your terminal shell to learn about
-all commands and their options.
+General help
+------------
+
+In order to get help for any ``cernopendata-client`` command, use the
+``--help`` option:
+
+.. code-block:: console
+
+    $ cernopendata-client --help
+    Usage: cernopendata-client [OPTIONS] COMMAND [ARGS]...
+
+    Options:
+    --help  Show this message and exit.
+
+    Commands:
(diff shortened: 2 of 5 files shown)


docs/cliapi.rst
```diff
@@ -6,6 +6,6 @@ CLI API
 Please use ``cernopendata-client --help`` in your terminal shell to learn about
 all commands and their options.
 
-.. click:: cernopendata_client.cli:cernopendata_client
```
tiborsimko, line 9: Now that #78 is merged, shall we try to simply plug back the old ReadTheDocs sphinx doc generation?

Please test locally using `readthedocs/build:latest` build image, because it may need to introduce some more `autodoc_mock_imports` in `docs/conf.py` with respect to both `pycurl` and `xrootd` packages.

ParthS007: done 👍

setup.py
```diff
@@ -32,7 +32,7 @@
     "docs": [
         "Sphinx>=1.4.4,<2.0",
         "sphinx-rtd-theme>=0.1.9",
-        "sphinx-click>=1.0.4",
+        "sphinx-click>=2.5.0",
```
tiborsimko, line 35: Could you at the same time amend the docstrings a bit so that the page looks prettier?  For example:

- [ ] `download-files` docstrings have two examples that are cutting lines; they could be shortened, or even simply removed, since the features like `–filter-range 1-2,3-4` are already showcased before

2020-11-06-153257_691x380_scrot

tiborsimko: [ ] Please add also suitable starting "Usage" section, listing all the commands, as we do for `reana-client` https://reana-client.readthedocs.io/en/latest/#cli-api

tiborsimko: - [ ] docs for `get-metadata` also contains one example that is going for more than one line.  You can shorten it.

docs/requirements.txt
```diff
@@ -1,5 +1 @@
-click>=7,<8
-requests==2.22.0
-Sphinx>=1.4.4,<2.0
-sphinx-click>=1.0.4
-sphinx-rtd-theme>=0.1.9
+-e .[docs]
```
tiborsimko, line 1: Please also add the usual copyright-related header:

```python
# This file is part of cernopendata-client.
#
# Copyright (C) 2019, 2020 CERN.
#
# cernopendata-client is free software; you can redistribute it and/or modify
# it under the terms of the GPLv3 license; see LICENSE file for more details.

-e .[docs]
```

ParthS007: Done 👍

docs/cliapi.rst
```diff
@@ -1,11 +1,29 @@
-.. _cliapi:
-
-CLI API
+CLI API
```
tiborsimko, line 1: Please remove trailing whitespace:

2020-11-06-160557_2560x1440_scrot

cernopendata_client/cli.py
```diff
@@ -80,7 +80,6 @@ def get_metadata(server, recid, doi, title, output_value):
     \b
     Examples:
       $ cernopendata-client get-metadata --recid 1
-      $ cernopendata-client get-metadata --recid 1 --output-value system_details.global_tag
```
tiborsimko, line 83: Here it is good to keep the line, since it is not demo'ed before... So just use something shorter, `--output-value title` for example.

docs/cliapi.rst
ParthS007: Done 👍

cernopendata_client/cli.py
ParthS007: Done 👍

tiborsimko: ... or `--output-value doi` if even more shortness is needed :wink:

tiborsimko (approved):

PR #81, opened 2020-11-06
docs: mention xrootd and pycurl installation options

### LINKED ISSUE #80: docs: mention xrootd and pycurl installation options
Following up #79, it would be good to do the following:

- [x] 1) Mention in https://cernopendata-client.readthedocs.io/en/latest/installation.html that people can specify two installation options, `pycurl` and `xrootd`, with a brief description what each one means.

- [x] 2) Currently, we actually use `xroot` name in `setup.py`, but I think we should use `xrootd` name because this is the official protocol name ([XRootD](https://xrootd.slac.stanford.edu/)).

```
    "xroot": ["xrootdpyfs>=0.2"],
``` 

- [x] 3) For consistency, let's also amend everywhere in our code base the terms like `--protocol root` to use `--protocol xrootd`, so that we say everywhere "xrootd" both in the code and the documentation.

ParthS007:
closes #80

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -19,6 +19,28 @@ You can also install it into a new virtual environment:
    $ source ~/.virtualenvs/cernopendata-client/bin/activate
    $ pip install cernopendata-client
 
+If you would like to use a more performing **pycurl** library for downloading data, please add the `pycurl` flavour while installing `cernopendata-client`:
+
+.. code-block:: console
+
+   $ pip install cernopendata-client[pycurl]
+
+Note that you need to have `curl <https://curl.se/>`_ and `openssl <https://www.openssl.org/>`_ set up on your system.
+
+Additionally, if you want to use an even more powerful **XRootD** protocol for downloading data, please add the `xrootd` flavour when installing `cernopendata-client`:
+
+.. code-block:: console
+
+   $ pip install cernopendata-client[xrootd]
+
+Note that you need to have `xrootd <https://xrootd.slac.stanford.edu/>`_ set up on your system.
+
+Finally, note that you can combine both flavours, if you wish to have both capabilities:
+
+.. code-block:: console
+
+   $ pip install cernopendata-client[pycurl,xrootd]
+
 Docker
 ------
 
--- a/cernopendata_client/cli.py
+++ b/cernopendata_client/cli.py
@@ -114,7 +114,7 @@ def get_metadata(server, recid, doi, title, output_value):
 @click.option(
     "--protocol",
     default="http",
-    type=click.Choice(["http", "root"]),
+    type=click.Choice(["http", "xrootd"]),
     help="Protocol to be used in links.",
 )
 @click.option(
@@ -136,7 +136,7 @@ def get_file_locations(server, recid, doi, title, protocol, expand):
     \b
     Examples:
       $ cernopendata-client get-file-locations --recid 5500
-      $ cernopendata-client get-file-locations --recid 5500 --protocol root
+      $ cernopendata-client get-file-locations --recid 5500 --protocol xrootd
     """
     validate_server(server)
     if recid is not None:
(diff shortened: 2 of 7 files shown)


docs/installation.rst
```diff
@@ -19,6 +19,27 @@ You can also install it into a new virtual environment:
    $ source ~/.virtualenvs/cernopendata-client/bin/activate
    $ pip install cernopendata-client
 
+In order to use Pycurl / xrootd related functionality, additional
+dependencies are required.
+
+The ``pycurl`` "extra" install flavor
```
tiborsimko, line 25: Instead of introducing new sections (1.2 and 1.3), this should be living as part of the 1.1 section.

Hence you can change formatting to something like:

This will use Python **requests** library for data download.

If you would like to use a more performing **pycurl** library for downloading data, please add the `pycurl` flavour while installing `cernopendata-client`:

```
$ pip install cernopendata-client[pycurl]
```

Note that you need to have curl (..link...) and openssl (...link...) set up on your system.

Additionally, if you want to use an even more powerful **XRootD** protocol for downloading data, please add the `xrootd` flavour when installing `cernopendata-client`:

```
$ pip install cernopendata-client[xrootd]
```

Note that you need to have XRootD (...link...) set up on your system.

Finally, note that you can combine both flavours:

```
$ pip install cernopendata-client[pycurl,xrootd]
```

if you wish to have both capabilities.

docs/installation.rst
```diff
@@ -19,6 +19,27 @@ You can also install it into a new virtual environment:
...
+-------------------------------------
+
+If you're installing via ``pip`` (recommended), you should be able to get the
+optional Python package requirements by changing your installation to refer to
+``cernopendata-client[pycurl]`` (from simply ``cernopendata-client``), e.g.::
+
+    pip install "cernopendata-client[pycurl]"
```
tiborsimko, line 32: Needs leading `$ ...` sign to indicate shell, so that it would go nicely with the other examples mentioned on that page.

(But see also the content changes suggested above.)

docs/installation.rst
```diff
@@ -19,6 +19,28 @@ You can also install it into a new virtual environment:
    $ source ~/.virtualenvs/cernopendata-client/bin/activate
    $ pip install cernopendata-client
 
+If you would like to use a more performing **pycurl** library for downloading data, please add the `pycurl` flavour while installing `cernopendata-client`:
+
+.. code-block:: console
+
+   $ pip install cernopendata-client[pycurl]
+
+Note that you need to have `curl <https://curl.se/download.html>`_ and `openssl <https://www.openssl.org/source/>`_ set up on your system.
```
tiborsimko, line 28: Perhaps just link to the homepages (rather than download pages) of the three tools?  

Chances are people have installed curl andd openssl already, so they may rather be interested to see what this XRootD stuff is all about, so rather seeing the home page instead of the download page.

(And nobody would use curl download to install it; people would use their package manager.)

P.S. While amending this, can you please look at the py3.7 test CI failure? Why did it fail?

docs/installation.rst
ParthS007: Done 👍

docs/installation.rst
ParthS007: Done 👍

docs/installation.rst
```diff
@@ -19,6 +19,28 @@ You can also install it into a new virtual environment:
...
+
+Additionally, if you want to use an even more powerful **XRootD** protocol for downloading data, please add the `xrootd` flavour when installing `cernopendata-client`:
+
+.. code-block:: console
+
+   $ pip install cernopendata-client[xrootd]
+
+Note that you need to have `xrootd <https://xrootd.slac.stanford.edu/>`_ set up on your system.
+
+Finally, note that you can combine both flavours, if you wish to have both capabilities.:
```
tiborsimko, line 38: `capabilities:`  (i.e. no dot before punctuation)

P.S. py37 test still fails?

ParthS007: Test is failing because the checksum is not matching in `test_download_files_http_pycurl/requests`. It is working fine on my machine. Looking into it.

docs/installation.rst
ParthS007: Done 👍

tiborsimko (approved):

PR #86, opened 2020-11-10
printer: simplify usage and standardise colouring

tiborsimko:
Simplifies usage of display_message() by keeping only one argument to
specify info/note/progress/error message types.

Tunes down the output colouring to resemble more faithfully the
practices from `yay` or `brew`, making the output more suitable for
various terminal colour schemes.

Pins `mock` dependency to version 3 so that Python-2.7 tests would
pass locally. (`tox -e py27`)


ParthS007 (approved):
Thanks for the PR. Tested locally. Works for me 🎉

PR #88, opened 2020-11-12
protocol: add 'https' option

### LINKED ISSUE #84: protocol: add 'https' option
We are deploying support for HTTPS for CERN Open Data portal, e.g. see https://opendata-dev.cern.ch

It would be therefore good to allow `https` protocol everywhere, so that we shall offer three protocols: `http`, `https`, and `xrootd`. 

We should amend the checker etc and also mention this in the documentation.

ParthS007:
closes #84

### DIFF
--- a/cernopendata_client/downloader.py
+++ b/cernopendata_client/downloader.py
@@ -48,7 +48,7 @@ def show_download_progress(
 
 def download_single_file(path=None, file_location=None, protocol=None):
     """Download single file."""
-    if protocol == "http":
+    if protocol in ["http", "https"]:
         file_name = file_location.split("/")[-1]
         file_dest = path + "/" + file_name
         with open(file_dest, "wb") as f:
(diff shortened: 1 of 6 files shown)


cernopendata_client/downloader.py
```diff
@@ -87,7 +87,7 @@ def download_single_file(path=None, file_location=None, protocol=None):
         if not xrootd_available:
             display_message(
                 msg_type="error",
-                msg="xrootd is not installed on system. Please use the 'http' protocol instead.",
+                msg="xrootd is not installed on system. Please use the 'http'/'https' protocol instead.",
```
tiborsimko, line 90: Please use the 'http' or 'https' protocol instead.

cernopendata_client/searcher.py
```diff
@@ -171,7 +171,7 @@ def get_files_list(server=None, record_json=None, protocol=None, expand=None):
 
     :param server: CERN Open Data server to query
     :param record_json: Record content in JSON
-    :protocol: Protocol to be used in links(http/xrootd)
+    :protocol: Protocol to be used in links(http/https/xrootd)
```
tiborsimko, line 174: Protocol to be used for links (http|https|xrootd)

Better to use `|` than `/`.

tiborsimko (commented):
Works nicely :+1: Left small cosmetic comments

tiborsimko (approved):
All the tests pass locally.

```
==== 56 passed in 16.41s ====
```

PR #90, opened 2020-11-13
Simplify run tests shellscript

ParthS007:


run-tests.sh
```diff
@@ -42,6 +42,16 @@ check_pytest () {
     python setup.py test
 }
 
+if [ $# -eq 0 ]; then
+    check_script
```
tiborsimko, line 46: You can also amend the names so that they would appear the same, i.e. generally make them lowercase.

E.g. in REANA we have `lint-black`, here we have `Lint-black`...

tiborsimko (approved):

PR #95, opened 2020-11-17
docker: XRootD client and build simplification

tiborsimko:
Installs XRootD command-line client.

Simplifies build instructions and removes unnecessary packages to make
the image slimmer.

Improves Dockerfile linting and adds a new CI check.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,8 +7,8 @@ on:
         branches: master
 
 jobs:
-    lint-Shellcheck:
-        runs-on: ubuntu-latest
+    lint-shellcheck:
+        runs-on: ubuntu-18.04
         steps:
           - uses: actions/checkout@v2
 
@@ -18,7 +18,7 @@ jobs:
               ./run-tests.sh --check-shellscript
 
     lint-black:
-        runs-on: ubuntu-latest
+        runs-on: ubuntu-18.04
         steps:
           - uses: actions/checkout@v2
 
@@ -33,7 +33,7 @@ jobs:
               ./run-tests.sh --check-black
 
     lint-pydocstyle:
-      runs-on: ubuntu-latest
+      runs-on: ubuntu-18.04
       steps:
         - uses: actions/checkout@v2
 
@@ -48,7 +48,7 @@ jobs:
             ./run-tests.sh --check-pydocstyle
 
     lint-flake8:
-      runs-on: ubuntu-latest
+      runs-on: ubuntu-18.04
       steps:
         - uses: actions/checkout@v2
 
@@ -63,7 +63,7 @@ jobs:
             ./run-tests.sh --check-flake8
 
     lint-check-manifest:
-      runs-on: ubuntu-latest
+      runs-on: ubuntu-18.04
       steps:
         - uses: actions/checkout@v2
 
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -77,6 +77,14 @@ jobs:
             pip install check-manifest
             ./run-tests.sh --check-manifest
 
+    lint-dockerfile:
+      runs-on: ubuntu-latest
+      steps:
+        - uses: actions/checkout@v2
+
+        - name: Lint Dockerfile
```
ParthS007, line 85:
```suggestion
        - name: Run hadolint
```

run-tests.sh
```diff
@@ -33,6 +33,14 @@ check_manifest () {
     check-manifest
 }
 
+check_dockerfile () {
+    hadolint Dockerfile
+}
+
+check_dockerfile () {
```
ParthS007, line 40:
```suggestion
check_dockerbuild () {
```

ParthS007 (commented):
We can keep a `hadolint.yaml` similar to `codecov.yaml` instead of adding inline config in `Dockerfile`.

wdyt?

tiborsimko:
> We can keep a `hadolint.yaml` similar to `codecov.yaml` instead of adding inline config in `Dockerfile`.
> 
> wdyt?

In REANA we went for having this inline, which makes things simpler.  Also, it is basically similar to inline `# noqa: F841` that we use for Python...

.github/workflows/ci.yml
tiborsimko: I have used the REANA term, but indeed here we use more verbose "Check compliance..." so I have changed this to "Check Dockerfile compliance.".

run-tests.sh
```diff
@@ -33,6 +33,14 @@ check_manifest () {
     check-manifest
 }
 
+check_dockerfile () {
+    hadolint Dockerfile
```
audrium, line 37: Maybe here it would also be better to use `hadolint` through `docker` (just like in `ci`). In that case it would also work for those who will not install `hadolint` locally

tiborsimko: Yes, we could do that. 

OT1H, it has a certain slowdown cost (0.26 sec vs 1.71 sec):

```console
$ time hadolint Dockerfile
hadolint Dockerfile  0.26s user 0.00s system 100% cpu 0.259 total
$ time docker run -i --rm hadolint/hadolint < Dockerfile
docker run -i --rm hadolint/hadolint < Dockerfile  0.03s user 0.04s system 4% cpu 1.707 total
```

So dunno. It may be acceptable, but if all developers install hadolint (e.g. system packages, or via Homebrew, or via Nix), we could also keep it "naked".  Anyway, the most important is to plug it into our editors.

OTOH, using docker here could mean that we can use `run-tests.sh --check-dockerfile` in GA CI to invoke the same recipe, so that we don't have to write it twice. Which is a good thing to do.

run-tests.sh
```diff
@@ -33,6 +33,15 @@ check_manifest () {
     check-manifest
 }
 
+check_dockerfile () {
+    docker run -i --rm hadolint/hadolint < Dockerfile
+}
+
+check_dockerbuild () {
```
audrium, line 40: Very minor detail: in `reana` this line is named `check_docker_build`. To make things consistent, which name is better and to which one we should stick to?

tiborsimko: It can go either way: 
- I had preferred `--check-dockerbuild` over `--check-docker-build` in one word so that it would be harmonious with `--check-dockerfile`. Basically, all the commands are being named `--check-something` where "something" is always one word, not sometimes one word sometimes two words.  
- But there is also a good point (=readability) behind going for `--check-docker-build`, `--check-js-tests`, etc...

tiborsimko: Since you have more PRs already done, I can simply amend towards `docker-build` with hyphen in this one :wink:

ParthS007 (approved):

PR #96, opened 2020-11-18
cli: Improve --verbose functionality

### LINKED ISSUE #94: cernopendata-client get-file-locations --recid 282 --protocol xrootd --verbose
This works:

```console
$ cernopendata-client get-file-locations --recid 282 --protocol xrootd | head -3
root://eospublic.cern.ch//eos/opendata/cms/Run2011B/Photon/AOD/12Oct2013-v1/00000/2A227E10-C949-E311-B033-003048FEAF50.root
root://eospublic.cern.ch//eos/opendata/cms/Run2011B/Photon/AOD/12Oct2013-v1/10000/0002E1AB-5A40-E311-9B07-0025901AF6E6.root
root://eospublic.cern.ch//eos/opendata/cms/Run2011B/Photon/AOD/12Oct2013-v1/10000/02C84E11-6040-E311-8B9A-C86000151BEC.root
```

This does not:

```console
$ cernopendata-client get-file-locations --recid 282 --protocol xrootd --verbose | head -3
root://eospublic.cern.ch//eos/opendata/cms/Run2011B/Photon/AOD/12Oct2013-v1/file-indexes/CMS_Run2011B_Photon_AOD_12Oct2013-v1_00000_file_index.json     269     sha1:5800af77c12d31bb76ef138d0b68cd6901facd9a
root://eospublic.cern.ch//eos/opendata/cms/Run2011B/Photon/AOD/12Oct2013-v1/file-indexes/CMS_Run2011B_Photon_AOD_12Oct2013-v1_10000_file_index.json     18615   sha1:5d91605740704cfba546e4766c67c8733d8de0c5
root://eospublic.cern.ch//eos/opendata/cms/Run2011B/Photon/AOD/12Oct2013-v1/file-indexes/CMS_Run2011B_Photon_AOD_12Oct2013-v1_20000_file_index.json     266736  sha1:70410acbcc711592115a14866e7485739622d188
```

In other words, the `--expand/--no-expand` command-line option is not respected when one uses `--verbose`.

ParthS007:
closes #94

### DIFF
--- a/cernopendata_client/searcher.py
+++ b/cernopendata_client/searcher.py
@@ -182,41 +182,43 @@ def get_files_list(
     :return: List of files list
     :rtype: list
     """
-    if verbose:
+    files_list = []
+    for file_ in record_json["metadata"]["files"]:
+        files_list.append((file_["uri"], file_["size"], file_["checksum"]))
+    if expand:
+        # let's unwind file indexes
+        files_list_expanded = []
+        for file_ in files_list:
+            if file_[0].endswith("_file_index.json"):
+                url_file = file_[0].replace(SERVER_ROOT_URI, server)
+                json_files = requests.get(url_file).json()
+                for file_ in json_files:
+                    files_list_expanded.append(
+                        (
+                            file_["uri"],
+                            file_["size"],
+                            file_["checksum"],
+                        )
+                    )
+            elif file_[0].endswith("_file_index.txt"):
+                pass
+            else:
+                files_list_expanded.append(file_)
+        files_list = files_list_expanded
+    if protocol == "http":
         files_list = [
-            (file["uri"], file["size"], file["checksum"])
-            for file in record_json["metadata"]["files"]
+            (file_[0].replace(SERVER_ROOT_URI, server), file_[1], file_[2])
+            for file_ in files_list
         ]
-        if protocol == "http":
-            files_list = [
-                (file_[0].replace(SERVER_ROOT_URI, server), file_[1], file_[2])
-                for file_ in files_list
-            ]
-    else:
-        files_list = [file["uri"] for file in record_json["metadata"]["files"]]
-        if expand:
-            # let's unwind file indexes
-            files_list_expanded = []
-            for file_ in files_list:
-                if file_.endswith("_file_index.txt"):
-                    url_file = file_.replace(SERVER_ROOT_URI, server)
-                    req = requests.get(url_file)
-                    for url_individual_file in req.text.split("\n"):
-                        if url_individual_file:
-                            files_list_expanded.append(url_individual_file)
-                elif file_.endswith("_file_index.json"):
-                    pass
-                else:
-                    files_list_expanded.append(file_)
-            files_list = files_list_expanded
-        if protocol == "http":
-            files_list = [
-                file_.replace(SERVER_ROOT_URI, server) for file_ in files_list
-            ]
-        elif protocol == "https":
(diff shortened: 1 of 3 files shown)


cernopendata_client/searcher.py
```diff
@@ -219,11 +233,27 @@ def get_files_list(
...
+            files_list = [
+                (file_[0].replace(SERVER_ROOT_URI, server), file_[1], file_[2])
+                for file_ in files_list
+            ]
+        else:
             files_list = [
                 file_.replace(SERVER_ROOT_URI, server) for file_ in files_list
             ]
-        elif protocol == "https":
+    elif protocol == "https":
```
tiborsimko, line 246: Can protocol be `https`? I think we have removed it fully, so it is only `http` or `xrootd`.  It is `server` that sets up HTTP or HTTPS. So we can probably simplify this code.

cernopendata_client/searcher.py
```diff
@@ -192,20 +192,34 @@ def get_files_list(
...
-                (file_[0].replace(SERVER_ROOT_URI, server), file_[1], file_[2])
-                for file_ in files_list
-            ]
-    else:
-        files_list = [file["uri"] for file in record_json["metadata"]["files"]]
-        if expand:
-            # let's unwind file indexes
+    files_list = []
+    for file_ in record_json["metadata"]["files"]:
+        if verbose:
```
tiborsimko, line 197: `verbose` is an output-level variable, so it should not really influence how we gather file information based on `expand`.
What about collecting always everything, i.e. uri/size/checksum triad, and use `verbose` only at the end to either output them or not? In this way the info-collecting part (using `expand`) and ouput-printing part (using `verbose`) will be more nicely separated.

cernopendata_client/searcher.py
ParthS007: Yes, It can be if `--server` is using `HTTPS`.

cernopendata_client/searcher.py
ParthS007: okay, sounds good 👍

cernopendata_client/searcher.py
tiborsimko: I mean do we even need to pass `protocol=https` to these functions in the searcher? For consistency, shouldn't we pass `protocol=http` and `server=https...` as we do in top-level CLI?  In this way, the Searcher expectations about server/protocol would fully comply with the CLI top-level notion of what server/protocol values might be... I'm wondering because inside Seacher we only use `protocol`, do no manipulation:

```console
$ git grep protocol cernopendata_client/searcher.py
cernopendata_client/searcher.py:167:    server=None, record_json=None, protocol=None, expand=None, verbose=N>
cernopendata_client/searcher.py:173:    :protocol: Protocol to be used in links [http,xrootd]
cernopendata_client/searcher.py:178:    :type protocol: str
cernopendata_client/searcher.py:208:    if protocol == "http":
cernopendata_client/searcher.py:213:    elif protocol == "https":
cernopendata_client/searcher.py:225:def get_file_info_remote(server, recid, protocol=None, filtered_files=No>
cernopendata_client/searcher.py:248:        if protocol == "http":
cernopendata_client/searcher.py:250:        elif protocol == "https":
```

The consistency between top-level CLI notion behind "protocol" and low-level Searcher notion behind "protocol", would avoid possible confusions later on.

Anyway, this could be also addressed later.

tiborsimko (approved):

PR #107, opened 2020-11-20
searcher: simplify protocol notion

### LINKED ISSUE #105: Consistency between top-level CLI notion behind "protocol" and low-level Searcher notion behind "protocol"
> I mean do we even need to pass `protocol=https` to these functions in the searcher? For consistency, shouldn't we pass `protocol=http` and `server=https...` as we do in top-level CLI?  In this way, the Searcher expectations about server/protocol would fully comply with the CLI top-level notion of what server/protocol values might be... I'm wondering because inside Seacher we only use `protocol`, do no manipulation:

```console
$ git grep protocol cernopendata_client/searcher.py
cernopendata_client/searcher.py:167:    server=None, record_json=None, protocol=None, expand=None, verbose=N>
cernopendata_client/searcher.py:173:    :protocol: Protocol to be used in links [http,xrootd]
cernopendata_client/searcher.py:178:    :type protocol: str
cernopendata_client/searcher.py:208:    if protocol == "http":
cernopendata_client/searcher.py:213:    elif protocol == "https":
cernopendata_client/searcher.py:225:def get_file_info_remote(server, recid, protocol=None, filtered_files=No>
cernopendata_client/searcher.py:248:        if protocol == "http":
cernopendata_client/searcher.py:250:        elif protocol == "https":
```

The consistency between top-level CLI notion behind "protocol" and low-level Searcher notion behind "protocol", would avoid possible confusions later on.

Anyway, this could be also addressed later.

_Originally posted by @tiborsimko in https://github.com/cernopendata/cernopendata-client/pull/96#discussion_r527010670_

ParthS007:
closes #105

### DIFF
--- a/cernopendata_client/searcher.py
+++ b/cernopendata_client/searcher.py
@@ -13,7 +13,7 @@ from __future__ import print_function
 import sys
 import requests
 
-from .config import SERVER_ROOT_URI, SERVER_HTTPS_URI
+from .config import SERVER_HTTP_URI, SERVER_ROOT_URI, SERVER_HTTPS_URI
 from .printer import display_message
 
 try:
@@ -182,6 +182,9 @@ def get_files_list(
     :return: List of files list
     :rtype: list
     """
+    searcher_protocol = protocol
+    if server != SERVER_HTTP_URI and searcher_protocol != "xrootd":
+        searcher_protocol = server.split(":")[0]
     files_list = []
     for file_ in record_json["metadata"]["files"]:
         files_list.append((file_["uri"], file_["size"], file_["checksum"]))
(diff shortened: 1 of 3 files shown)


cernopendata_client/searcher.py
```diff
@@ -205,12 +208,12 @@ def get_files_list(
...
                 files_list_expanded.append(file_)
         files_list = files_list_expanded
-    if protocol == "http":
+    if searcher_protocol == "http":
         files_list = [
             (file_[0].replace(SERVER_ROOT_URI, server), file_[1], file_[2])
             for file_ in files_list
         ]
-    elif protocol == "https":
+    elif searcher_protocol == "https":
```
tiborsimko, line 216: Here is one situation I found that is not working:

```console


$ cernopendata-client get-file-locations --recid 5500 --server https://opendata.cern.ch --protocol xrootd | head -3
https://opendata.cern.ch/eos/opendata/cms/software/HiggsExample20112012/BuildFile.xml
https://opendata.cern.ch/eos/opendata/cms/software/HiggsExample20112012/HiggsDemoAnalyzer.cc
https://opendata.cern.ch/eos/opendata/cms/software/HiggsExample20112012/List_indexfile.txt
```

The value of `protocol` is not respected when `server` starts with `https`.

Can you please amend and add a test case?
```

tiborsimko (approved):

Issue #99, opened 2020-11-19
downloader: resume interrupted downloads


tiborsimko:
When one download a big 20 GB file, something can go wrong along the way, or the user can sleep or restart their machines. When this happens and user re-issues `download-file` command, this will start download again from zero, event though local directory already has parts of the file. 

The goal of this issue to resume interrupted downloads from the last good state.

How to reproduce:

1. Run local CERN Open Data instance as follows:

```console
$ cd opendata.cern.ch
$ docker-compose build
$ docker-compose up
$ docker exec -i -t opendatacernch_web_1 ./scripts/populate-instance.sh --skip-records
$ docker exec -i -t opendatacernch_web_1 cernopendata fixtures records --mode insert-or-replace -f cernopendata/modules/fixtures/data/records/atlas-2020-exactly2lep.json
$ firefox http://localhost/record/15007
```

The test record contains a big 20 GB file.

Now start a download and interrupt after a while:

```console
$ cernopendata-client download-files --recid 15007 --server http://localhost
==> Downloading file 1 of 1
  -> File: ./15007/exactly2lep.zip
 
^C
$ ls -lh 15007/exactly2lep.zip
-rw-r--r-- 1 simko simko 15M Nov 19 11:40 15007/exactly2lep.zip
```

and resume download from scratch again:

```console
$ cernopendata-client download-files --recid 15007 --server http://localhost
```

The downloader should recognise already available `15007/exactly2lep.zip` file and should continue from there.

Note that the user can interrupt download any number of times until successful completion.

The behaviour could be configurable by a new `--resume` option, for example: 
- when the downloader sees there is no target file yet, it proceeds to download as usual;
- when the downloader sees the file, it checks it size and checksum, and
   - if the file is complete, it will say that there is nothing to download, that the file is here and is already verified
   - if the file is partial, it would check whether the user used `--resume` option, and if yes, then would continue from that point, and if no, it would ask the user whether resume or redownload is wanted.  
- Since we have only big files, I guess the resume behaviour could be the default though, and ask the user to do `rm ...` on the given file if the user would like *not* to do resume, but rather redownload. 

CC @katilp

ParthS007:
@tiborsimko 

> if the file is partial, it would check whether the user used --resume option, and if yes, then would continue from that point, and if no, it would ask the user whether the resume or redownload is wanted.

I have a couple of musings.

1. We will know if a file is partially downloaded when it is not matching the checksum and size from the remote file.

[...]

Issue #98, opened 2020-11-18
downloader: catch more error situations


tiborsimko:
Now that we are catching downloading errors via `--retry-limit` and `--retry-sleep` parameters thanks to #91, we can build upon this basis and enrich the error handling logic to catch more errors (such as Error 500 received from server, etc).

Example how to reproduce:

1. Run locally a CERN Open Data instance:

```console
$ cd opendata.cern.ch
$ docker-compose build
$ docker-compose up
$ docker exec -i -t opendatacernch_web_1 ./scripts/populate-instance.sh --skip-records
$ docker exec -i -t opendatacernch_web_1 cernopendata fixtures records --mode insert-or-replace -f cernopendata/modules/fixtures/data/records/cms-tools-higgsexample20112012.json
$ firefox http://localhost/record/5500
```

2. Launch downloading of files with big enough retry sleep time:

```console
$ cernopendata-client download-files --recid 5500 --verify  --server http://localhost --retry-sleep 30
```

3. While the download is running, bring down the web site.

Current behaviour of the client is to output Python traceback:

```console
==> Verifying file demoanalyzer_cfg_level4data.py...
  -> Expected size 3821, found 3821
  -> Expected checksum adler32:177b49c0, found adler32:177b49c0
==> Downloading file 10 of 11
  -> File: ./5500/mass4l_combine.pdf
Traceback (most recent call last):
  File ".../bin/cernopendata-client", line 8, in <module>
    sys.exit(cernopendata_client())
  File ".../lib/python3.8/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File ".../lib/python3.8/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File ".../lib/python3.8/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File ".../lib/python3.8/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File ".../lib/python3.8/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File ".../lib/python3.8/site-packages/cernopendata_client/cli.py", line 337, in download_files
    download_single_file(path=path, file_location=file_location, protocol=protocol)
  File ".../lib/python3.8/site-packages/cernopendata_client/downloader.py", line 109, in download_single_file
    c.perform()
[...]

PR #109, opened 2020-11-24
downloader: catch more error situations

ParthS007:
addresses #98 #99

### DIFF
--- a/cernopendata_client/searcher.py
+++ b/cernopendata_client/searcher.py
@@ -31,7 +31,7 @@ def verify_recid(server=None, recid=None):
     :type server: str
     :type recid: int
 
-    :return: Boolean after verifying the record id
+    :return: Boolean after verifying the record ID
     :rtype: bool
     """
     input_record_url = server + "/record/" + str(recid)
@@ -46,7 +46,7 @@ def verify_recid(server=None, recid=None):
         except Exception:
             display_message(
                 msg_type="error",
-                msg="The record id number you supplied is not valid.",
+                msg="The record ID number you supplied is not valid.",
             )
             sys.exit(1)
         return False
@@ -70,14 +70,14 @@ def get_recid_api(server=None, base_record_id=None):
     except Exception:
         display_message(
             msg_type="error",
-            msg="The record id number you supplied is not valid.",
+            msg="The record ID number you supplied is not valid.",
         )
         sys.exit(1)
     return record_api
 
 
 def get_recid(server=None, title=None, doi=None):
-    """Return record id by either title or doi.
+    """Return record ID by either title or doi.
 
     :param server: CERN Open Data server to query
     :param title: Record title
@@ -86,7 +86,7 @@ def get_recid(server=None, title=None, doi=None):
     :type title: str
     :type doi: str
 
-    :return: record id
+    :return: record ID
     :rtype: int
     """
     if title:
--- a/cernopendata_client/searcher.py
+++ b/cernopendata_client/searcher.py
@@ -43,7 +43,7 @@ def verify_recid(server=None, recid=None):
     else:
         try:
             input_record_url_check.raise_for_status()
-        except requests.HTTPError:
+        except Exception:
             display_message(
                 msg_type="error",
                 msg="The record id number you supplied is not valid.",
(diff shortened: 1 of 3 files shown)


cernopendata_client/searcher.py
```diff
@@ -96,7 +103,7 @@ def get_recid(server=None, title=None, doi=None):
     response_json = response.json()
     try:
         response.raise_for_status()
-    except requests.HTTPError as e:
+    except Exception as e:
```
tiborsimko, line 106: It would be nice to list all the exceptions instead of one single catch all.

BTW the scenario described in #98 does not for me fully with this PR.  I get:

```
ConnectionResetError: [Errno 104] Connection reset by peer
...
During handling of the above exception, another exception occurred:
...
urllib3.exceptions.ProtocolError: ('Connection aborted.', ConnectionResetError(104, 'Connection reset by peer'))
...
During handling of the above exception, another exception occurred:
...
requests.exceptions.ConnectionError: ('Connection aborted.', ConnectionResetError(104, 'Connection reset by peer'))
```

Is this improved with sequel PRs? Shall we address them in one go?

ParthS007: Yes, This PR addresses #98 and will be completed in the sequel of PRs as we discussed.

tiborsimko: So let's perhaps squash them together, it'll be easier to test that way!

tiborsimko:
Repetitive download leads to an issue.  Here's a test case. The current behaviour is:

```console
$ git clean -d -ff -x
$ cernopendata-client download-files --recid 5500 --verify
==> Downloading file 1 of 11
  -> File: ./5500/BuildFile.xml
  -> Progress: 0/0 KiB (100%)
==> Verifying file BuildFile.xml...
  -> Expected size 305, found 305
  -> Expected checksum adler32:ff63668a, found adler32:ff63668a
...
==> Downloading file 11 of 11
  -> File: ./5500/mass4l_combine.png
  -> Progress: 90/90 KiB (100%)
==> Verifying file mass4l_combine.png...
  -> Expected size 93152, found 93152
  -> Expected checksum adler32:62e0c299, found adler32:62e0c299
==> Success!
$ $ cernopendata-client download-files --recid 5500 --verify

==> Downloading file 1 of 11
  -> File {} is complete. Skip download.
```

Expected behaviour:
- do not stop after 1st file
- say properly which file is OK (f-string), something like:
  "File ./5500/foo.py is up to date; skipping download
- note that for xrootd protocol, the files are always re-downloaded, there is no check. It would be good to harmonise the behaviour, e.g. always-re-download or always-skip for both HTTP and XRootD protocols

ParthS007:
- Pushed more changes, now you can check resuming-downloading with the `requests` library. 

- Added a couple of functions separating the functionalities, names can be changed.

- I have harmonized the `re-download` now files will `always-re-download` if they are all present for now. We can change it once we are done with adding resuming downloads for all libs.

cernopendata_client/searcher.py
```diff
@@ -43,7 +43,7 @@ def verify_recid(server=None, recid=None):
     else:
         try:
             input_record_url_check.raise_for_status()
-        except requests.HTTPError:
+        except Exception:
             display_message(
                 msg_type="error",
                 msg="The record id number you supplied is not valid.",
```
tiborsimko, line 49: 'record id' -> 'record ID'

Can you please update this part, even if it is not part of the PR?  Also in other places:

```console
$ rg 'record id'
cernopendata_client/searcher.py
34:    :return: Boolean after verifying the record id
49:                msg="The record id number you supplied is not valid.",
73:            msg="The record id number you supplied is not valid.",
80:    """Return record id by either title or doi.
89:    :return: record id

cernopendata_client/validator.py
29:            msg="You must supply a record id number as an " "input using -r flag.",
```

Could be a separate commit.

cernopendata_client/searcher.py
tiborsimko: The general `except` catch-all is not the best paradigm, as it can mask some error situations, but it is OK for now for partial addressing...

tiborsimko (approved):

tiborsimko:
Tests of the "resume interrupted download" feature passed locally with `requests` downloaded, and otherwise in general:

```
  py27: commands succeeded
  py36: commands succeeded
  py37: commands succeeded
  py38: commands succeeded
  py39: commands succeeded
  congratulations :)
```

PR #110, opened 2020-11-25
searcher: add KeyError handling and test case

### LINKED ISSUE #108: cernopendata-client get-metadata --recid 550 --output-value title
Some records do not have files, and there is a traceback for various commands, e.g.

- getting some metadata value:

```console
$ cernopendata-client get-metadata --recid 550 --output-value title
...
  File ".../cernopendata_client/searcher.py", line 157, in get_record_as_json
    if record_json["metadata"]["files"]:
KeyError: 'files'
```

- getting list of files:

```console
$ cernopendata-client get-file-locations --recid 550
...
  File ".../cernopendata_client/searcher.py", line 157, in get_record_as_json
    if record_json["metadata"]["files"]:
KeyError: 'files'
```

We should:
- amend the code so that no traceback is outputted, rather empty list of files;
- write a  test case to make sure this situation is handled correctly.

ParthS007:
closes #108

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -102,9 +102,7 @@ jobs:
             sudo apt-get install libgnutls28-dev
 
         - name: Install Python dependencies
-          run: |
-            pip install --upgrade pip setuptools py
-            pip install -e .[docs]
+          run: pip install -e .[docs]
 
         - name: Run Sphinx documentation with doctests
           run: ./run-tests.sh --check-sphinx
@@ -131,7 +129,6 @@ jobs:
 
           - name: Install Python dependencies
             run: |
-              pip install --upgrade pip setuptools py
               pip install wheel
               pip install -e .[all]
 
--- a/cernopendata_client/searcher.py
+++ b/cernopendata_client/searcher.py
@@ -154,12 +154,15 @@ def get_record_as_json(server=None, recid=None, doi=None, title=None):
     record_json = record_api.json()
     if "_files" in record_json["metadata"]:
         del record_json["metadata"]["_files"]
-    if record_json["metadata"]["files"]:
-        for field in record_json["metadata"]["files"]:
-            if "bucket" in field:
-                del field["bucket"]
-            if "version_id" in field:
-                del field["version_id"]
+    try:
+        if record_json["metadata"]["files"]:
+            for field in record_json["metadata"]["files"]:
+                if "bucket" in field:
+                    del field["bucket"]
+                if "version_id" in field:
+                    del field["version_id"]
+    except KeyError:
+        record_json["metadata"]["files"] = []
     return record_json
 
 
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -131,7 +131,7 @@ jobs:
 
           - name: Install Python dependencies
             run: |
-              pip install --upgrade pip setuptools py
+              pip install --upgrade setuptools py
```
tiborsimko, line 134: What about removing also `setuptools` update? Cannot we simply use the one coming with the system?

Also, is `py` needed there?

Perhaps remove the whole `--upgrade` line...

tiborsimko: Also, we have the same `pip --upgrade` situation above, see line 106. That might be removed too.

cernopendata_client/searcher.py
```diff
@@ -154,12 +154,15 @@ def get_record_as_json(server=None, recid=None, doi=None, title=None):
     record_json = record_api.json()
     if "_files" in record_json["metadata"]:
         del record_json["metadata"]["_files"]
-    if record_json["metadata"]["files"]:
-        for field in record_json["metadata"]["files"]:
-            if "bucket" in field:
-                del field["bucket"]
-            if "version_id" in field:
-                del field["version_id"]
+    try:
```
tiborsimko, line 157: It works OK. However, can you please fix similar problems in the other commands where "flies" may not exist? For example:

```console
$ cernopendata-client get-file-locations --recid 550
...
  File ".../cernopendata_client/searcher.py", line 189, in get_files_list
    for file_ in record_json["metadata"]["files"]:
KeyError: 'files'
```

Generally speaking, instead of code like:

```python
for file_ in record_json["metadata"]["files"]:
    pass
```

it is better to always do:

```python
for file_ in record_json["metadata"].get("files", []):
    pass
```

for all the optional fields.

See `rg -C 1 '\["files"\]'` for all occurrences where this may be needed.

tiborsimko: P.S. If you do this, then the `try/except` wouldn't be needed, and we could simply do something like:

```python
if record_json["metadata"].get("files", []):
    for field in record_json["metadata"].get("files", []):
        ...
```

so that KeyError wouldn't theoretically mask more situations of not existing keys (files, bucket, version_id) in case the if/for loop code could happen to get longer and longer in the future.

tiborsimko: Oh, I must have been on a different code, this works now:

```console
$ cernopendata-client get-file-locations --recid 550  --protocol xrootd
```

So if you wish we can merge "as is" without amending the key business...

ParthS007: Opened another issue: https://github.com/cernopendata/cernopendata-client/issues/112 for failing python test.

tiborsimko (approved):

tiborsimko:
(FYI tests passed locally; merging)

PR #113, opened 2020-12-18
searcher: Fix JSONdecodeError in unwinding file indexes

### LINKED ISSUE #112: searcher: JSON Decode error for `--verbose`
From [MM chat](https://mattermost.web.cern.ch/it-dep/pl/kkmabo3643y7ty6jk16brt7mpw):
> I guess this is all due to EOS gateway, we can attack the root cause via handling `/files/...` URIs. 
   Just to see if it works better.
   Example:`http://opendata.cern.ch/eos/opendata/cms/Run2010B/BTau/AOD/Apr21ReReco-v1/file-indexes/CMS_Run2010B_BTau_AOD_Apr21ReReco-v1_0000_file_index.json`
  is also available as:
  `http://opendata.cern.ch/record/1/files/CMS_Run2010B_BTau_AOD_Apr21ReReco-v1_0000_file_index.json`

ParthS007:
closes #112

### DIFF
--- a/cernopendata_client/searcher.py
+++ b/cernopendata_client/searcher.py
@@ -203,8 +203,17 @@ def get_files_list(
         files_list_expanded = []
         for file_ in files_list:
             if file_[0].endswith("_file_index.json"):
-                url_file = file_[0].replace(SERVER_ROOT_URI, server)
-                json_files = requests.get(url_file).json()
+                try:
+                    url_file = "{}/record/{}/files/{}".format(
+                        server, str(record_json["id"]), file_[0].split("/")[-1]
+                    )
+                    json_files = requests.get(url_file).json()
+                except Exception:
+                    display_message(
+                        msg_type="error",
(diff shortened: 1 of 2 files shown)


cernopendata_client/searcher.py
```diff
@@ -203,8 +203,17 @@ def get_files_list(
         files_list_expanded = []
         for file_ in files_list:
             if file_[0].endswith("_file_index.json"):
-                url_file = file_[0].replace(SERVER_ROOT_URI, server)
-                json_files = requests.get(url_file).json()
+                try:
+                    url_file = "{}/record/{}/files/{}".format(
```
tiborsimko, line 207: This fixes the problem at hand.

That said, I'm wondering whether we shouldn't introduce at this point in time already the distinction between EOSPUBLIC locations and `/record/NNN/files/foo.root` locations. Since this topic will touch other parts, I'll probably make another ticket about this.

tiborsimko (approved):

PR #117, opened 2021-01-04
downloader: resume interrupted downloads

ParthS007:
Add resume downloads for `pycurl` library
addresses #99


tiborsimko (approved):
Works well :+1: The refactoring of underlying downloader helpers can come later.

PR #118, opened 2021-01-05
global: refactor downloader code

ParthS007:
global: refactor downloader code
cli: add --download-lib option in download-files

### DIFF
--- a/cernopendata_client/cli.py
+++ b/cernopendata_client/cli.py
@@ -243,6 +243,14 @@ def get_file_locations(server, recid, doi, title, protocol, expand, verbose):
         DOWNLOAD_RETRY_SLEEP
     ),
 )
+@click.option(
+    "--download-engine",
+    "download_engine",
+    type=click.Choice(["requests", "pycurl", "xrootdpyfs"]),
+    help="Download engine to use when downloading files."
+    "The available values are 'requests', 'pycurl', 'xrootdpyfs'."
+    "[default=requests (for HTTP protocol), xrootdpyfs (for XRootD protocol)]",
+)
 def download_files(
     server,
     recid,
@@ -257,6 +265,7 @@ def download_files(
     verify,
     retry_limit,
     retry_sleep,
+    download_engine,
 ):
     """Download data files belonging to a record.
 
@@ -330,6 +339,11 @@ def download_files(
                 msg_type="error",
                 msg="Creation of the directory {} failed".format(path),
             )
+    if not download_engine:
+        if protocol.startswith("http"):
+            download_engine = "requests"
+        elif protocol == "xrootd":
+            download_engine = "xrootdpyfs"
     for file_location in download_file_locations:
         display_message(
             msg_type="info",
@@ -337,7 +351,12 @@ def download_files(
                 download_file_locations.index(file_location) + 1, total_files
             ),
         )
-        download_single_file(path=path, file_location=file_location, protocol=protocol)
+        download_single_file(
+            path=path,
+            file_location=file_location,
+            protocol=protocol,
+            download_engine=download_engine,
+        )
         check_error(
             path=path,
             file_location=file_location,
--- a/cernopendata_client/downloader.py
+++ b/cernopendata_client/downloader.py
@@ -8,11 +8,18 @@
 
 """cernopendata-client file downloading related utilities."""
 
+from __future__ import print_function
 import sys
 import os
 import re
 import time
-import requests
+
+try:
+    import requests
+
+    requests_available = True
+except ImportError:
+    requests_available = False
 
 try:
     import pycurl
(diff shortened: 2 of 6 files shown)


cernopendata_client/downloader_class.py
```diff
@@ -0,0 +1,115 @@
...
+    pycurl_available = True
+except ImportError:
+    pycurl_available = False
+
+
+from .printer import display_message
+
+
+class Downloader(object):
+    """Downloader class for managing pycurl and requests related utilities functions."""
```
tiborsimko, line 32: If we go for the object-oriented way, what about bringing it further:

- make Downloader class abstract, with common methods for showing progress etc, and with abstract method download() that will get implemented in each subclass
- create DownloaderPycurl and DownloaderRequests classes implementing download() function using each concrete downloader

This will also leave space for creating new classes:
- DownloaderXRootD class, using xrootdpyfs; this is currently enough. But later we could have another class using vanilla xrootd client (as opposed to xrootdpyfs).

Note: OT1H, the common Downloader() class could take care of both HTTP and XRootD protocols, if it makes sense to have some common functions, such as progress reporting or argument massaging. OTOH, we could also make two distinct class hierarchies, DownloaderHttp (with its DownloaderHttpPycurl, DownloaderHttpRequests subclasses) and DownloaderXrootd, which could make separation cleaner.

cernopendata_client/cli.py
```diff
@@ -243,6 +243,13 @@ def get_file_locations(server, recid, doi, title, protocol, expand, verbose):
         DOWNLOAD_RETRY_SLEEP
     ),
 )
+@click.option(
+    "--download-lib",
+    "dl_lib",
```
tiborsimko, line 248: `dl_lib` is not very descriptive. What about keeping the same variable name as the CLI option name. This will make searchers etc easier.

`--download-lib` would be better spelled `--download-library`. But if we keep this option also for XRootD and any other future methods, what about simply calling it `--downloader`? Or perhaps `--downloader-engine`?

cernopendata_client/downloader_class.py
```diff
@@ -0,0 +1,115 @@
+# -*- coding: utf-8 -*-
+# This file is part of cernopendata-client.
```
tiborsimko, line 2: Do we need a new file? The class is relatively small, and if we introduce subclasses, the code will be nicely separated in that way.  So I guess it can stay in the same `downloader.py` file.  It would be simpler and more consistent with the other files present in the cernopendata-client package.

cernopendata_client/cli.py
ParthS007: `--downloader-engine` sounds good to me.

cernopendata_client/downloader_class.py
ParthS007: I created the new file just to have the class separated to just show how it will look like, will move it to `downloader.py`.

cernopendata_client/downloader_class.py
ParthS007: The option with making `Downloader` class abstract and implementing  `DownloaderPycurl` and `DownloaderRequests`, `DownloaderXRootD` looks good to me.

I am not sure how the progress or any other methods will get implemented in `Xrootd` case so making distinct class hierarchies, DownloaderHttp (with its DownloaderHttpPycurl, DownloaderHttpRequests subclasses) and DownloaderXrootd also makes sense.

Bit confused here for choosing any of these. 😄

ParthS007: After [MM](https://mattermost.web.cern.ch/it-dep/pl/y5zgyb3153gp7xmigce1fo6hne) discussion, proceeded with the second option.

cernopendata_client/downloader.py
```diff
@@ -34,15 +40,176 @@
...
+            except Exception:
+                display_message(
+                    msg_type="error",
+                    msg="Download error occured. Please try again.",
+                )
+                sys.exit(1)
+            c.close()
+
+
+class DownloaderXrootd:
```
tiborsimko, line 151: The new XRootD downloader does not work locally for me, the troubles are already with checking local file existence before attempting the first download... Here's a simple recipe:

```
$ git clean -d -ff -x
$ $ cernopendata-client download-files --recid 5500  --protocol xrootd
==> Downloading file 1 of 11
Traceback (most recent call last):
  File ".../cernopendata-client/bin/cernopendata-client", line 8, in <module>
    sys.exit(cernopendata_client())
  File ".../cernopendata-client/lib/python3.9/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File ".../cernopendata-client/lib/python3.9/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File ".../cernopendata-client/lib/python3.9/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File ".../cernopendata-client/lib/python3.9/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File ".../cernopendata-client/lib/python3.9/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File ".../cernopendata-client/lib/python3.9/site-packages/cernopendata_client/cli.py", line 354, in download_files
    check_error(
  File ".../cernopendata-client/lib/python3.9/site-packages/cernopendata_client/downloader.py", line 211, in check_error
    "size": os.path.getsize(file_dest),
  File "/usr/lib64/python3.9/genericpath.py", line 50, in getsize
    return os.stat(filename).st_size
FileNotFoundError: [Errno 2] No such file or directory: '5500/BuildFile.xml'

```

tests/test_cli_download_files.py
```diff
@@ -123,6 +123,24 @@ def test_download_files_https_requests(mocker):
         os.remove(test_file)
 
 
+def test_download_files_download_engine(mocker):
```
tiborsimko, line 126: I see troubles with Python-2.7 locally. The `importlib-resources` was installed but is not Python-2.7 friendly, since it uses f-strings. You need to pin some package versions in `setup.py` conditionally, so that they would not be installed for Python-2.7.  Here is a test case to try out locally: 

```console
$ tox -e py27
...
  File "/home/tibor/private/project/opendata/src/cernopendata-client/.eggs/importlib_resources-0.0.0-py2.7.egg/importlib_resources/readers.py", line 77
    raise FileNotFoundError(f'{self} is not a file')
                                                  ^
...
  File "/home/tibor/private/project/opendata/src/cernopendata-client/.tox/py27/lib/python2.7/site-packages/pkg_resources/__init__.py", line 786, in resolve
    raise DistributionNotFound(req, requirers)
pkg_resources.DistributionNotFound: The 'importlib-resources>=1.0' distribution was not found and is required by virtualenv
ERROR: InvocationError for command /home/tibor/private/project/opendata/src/cernopendata-client/.tox/py27/bin/python setup.py test (exited with code 1)
```

cernopendata_client/cli.py
```diff
@@ -243,6 +243,13 @@ def get_file_locations(server, recid, doi, title, protocol, expand, verbose):
         DOWNLOAD_RETRY_SLEEP
     ),
 )
+@click.option(
+    "--download-engine",
+    "download_engine",
+    default="requests",
+    type=click.Choice(["requests", "pycurl"]),
+    help="Library to be used in downloading files [requests,pycurl, xrootd] [default=requests]",
```
tiborsimko, line 251: Please replace "Library by "Engine" here and elsewhere, since the option is called "engine".

cernopendata_client/downloader.py
```diff
@@ -179,41 +263,63 @@ def downloader_file_checker(file_location, file_dest):
...
+            "pycurl": pycurl_available,
+        }
+        if download_engine:
+            if not download_engine_map.get(download_engine):
                 display_message(
-                    msg_type="note",
-                    msg="File {} is incomplete. Resuming download.".format(
-                        file_name,
+                    msg_type="error",
+                    msg="{} is not installed on system. Please install it".format(
```
tiborsimko, line 294: Note that there is a traceback coming after the message. We should exit without showing traceback. Also, please put a full stop after the error message.

```console
$ cernopendata-client download-files --recid 5500 --download-engine=pycurl
==> Downloading file 1 of 11
==> ERROR: pycurl is not installed on system. Please install it
Traceback (most recent call last):
  File ".../cernopendata-client/bin/cernopendata-client", line 8, in <module>
    sys.exit(cernopendata_client())
  File ".../cernopendata-client/lib/python3.9/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File ".../cernopendata-client/lib/python3.9/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File ".../cernopendata-client/lib/python3.9/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File ".../cernopendata-client/lib/python3.9/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File ".../cernopendata-client/lib/python3.9/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File ".../cernopendata-client/lib/python3.9/site-packages/cernopendata_client/cli.py", line 348, in download_files
    download_single_file(
  File ".../cernopendata-client/lib/python3.9/site-packages/cernopendata_client/downloader.py", line 322, in download_single_file
    downloader.file_downloader()
  File ".../cernopendata-client/lib/python3.9/site-packages/cernopendata_client/downloader.py", line 132, in file_downloader
    c = pycurl.Curl()
NameError: name 'pycurl' is not defined
```

tiborsimko: Another observation on the same track: is protocol option being checked against download engine option?

```console
$ cernopendata-client download-files --recid 5500 --protocol xrootd --download-engine requests
==> Downloading file 1 of 11
```

One would expect an error message saying that protocol xrootd and engine requests are not compatible. 

Please add some checks and test cases for that.

tests/test_cli_download_files.py
ParthS007: I tested the `$ tox -e py27` in a `python2.7 env`. It worked fine without warnings.
[log.txt](https://github.com/cernopendata/cernopendata-client/files/5800942/log.txt)

cernopendata_client/cli.py
```diff
@@ -243,6 +243,12 @@ def get_file_locations(server, recid, doi, title, protocol, expand, verbose):
         DOWNLOAD_RETRY_SLEEP
     ),
 )
+@click.option(
+    "--download-engine",
```
tiborsimko, line 247: We are using `download-engine` but elsewhere in the code we have `downloader-engine`:

```
tests/test_cli_download_files.py
127:    """Test download_files() command with downloader-engine option."""
145:    """Test download_files() command with downloader-engine option and wrong protocol."""
157:    """Test download_files() command with downloader-engine option and wrong protocol."""
```

Let's standardise this.  

BTW what about the simplest version `downloader` so that people type `--downloader pycurl`?  Seems good to go nicely with class names...

cernopendata_client/cli.py
```diff
@@ -243,6 +243,12 @@ def get_file_locations(server, recid, doi, title, protocol, expand, verbose):
...
+    "download_engine",
+    type=click.Choice(["requests", "pycurl", "xrootdpyfs"]),
+    help="Engine to be used in downloading files [requests, pycurl, xrootdpyfs] [default=requests(for HTTP), xrootdpyfs(for XRootD)]",
```
tiborsimko, line 250: Download engine to use when downloading files. The available values are 'requests', 'pycurl', 'xrootdpyfs'. [default=requests (for HTTP protocol), xrootdpyfs (for XRootD protocol)]

docs/usage.rst
```diff
@@ -247,6 +247,22 @@ xrootd`` command-line option to use that protocol instead of HTTP/HTTPS:
       -> File: ./5500/mass4l_combine.png
     ==> Success!
 
+**Downloading engine**
```
tiborsimko, line 250:
```
*Select downloader engine*
```

to go better with the other subtitles that use the verb style such as "Filter by name..."

cernopendata_client/cli.py
ParthS007: I was thinking about `downloader` initially but then I thought `download-engine` sounds to be more specific.

docs/usage.rst
```diff
@@ -247,6 +247,22 @@ xrootd`` command-line option to use that protocol instead of HTTP/HTTPS:
...
+
+You can specify the download engine with ``--download-engine`` option. 
+
+- ``requests`` and ``pycurl`` are supported download engine for **HTTP** protocol.
```
tiborsimko, line 254: are two supported download engines for

(plural)

docs/usage.rst
```diff
@@ -247,6 +247,22 @@ xrootd`` command-line option to use that protocol instead of HTTP/HTTPS:
...
+- ``xrootdpyfs`` is supported download engine for **XRootD** protocol.
```
tiborsimko, line 255: is the only supported

docs/usage.rst
```diff
@@ -247,6 +247,22 @@ xrootd`` command-line option to use that protocol instead of HTTP/HTTPS:
...
+
+.. code-block:: console
+
+    $ cernopendata-client download-files --recid 5500 --filter-name BuildFile.xml --download-engine pycurl
+    ==> Downloading file 1 of 1
+      -> File: ./5500/BuildFile.xml
+      -> Progress: 0/0 KiB (100%)
+    ==> Success!
+
```
tiborsimko, line 264: remove extra empty line

tests/test_cli_download_files.py
tiborsimko: Solved: this was observed with OS native `python3-tox` package. Using `tox` from PyPI fixes the troubles.

cernopendata_client/downloader.py
```diff
@@ -179,74 +270,96 @@ def downloader_file_checker(file_location, file_dest):
...
-                )
-                downloader_pycurl(
-                    file_location, path, file_size_offline=file_size_offline, mode="ab"
-                )
-            else:
-                downloader_pycurl(file_location, path, mode="wb")
+        if download_engine not in download_engine_protocol_http_map:
+            display_message(
+                msg_type="error",
+                msg="{} is not compatible with {} protocol. Please use requests/pycurl engine.".format(
```
tiborsimko, line 312: "Please use requests/pycurl" could be misunderstood.  Better say "Please use 'requests' or 'pycurl'."

tiborsimko: BTW, in order to have this extensible, perhaps you can have a `config.py` variable which protocol supports which download engine options, and print the message dynamically from that config variable?

(E.g. this will come handy when we shall add vanilla 'xrdcp' as opposed to 'xrootdpyfs'.)

setup.py
```diff
@@ -27,6 +27,7 @@
     "pytest>=2.8.0",
     'pytest-mock>=2.0,<3.0 ; python_version=="2.7"',
     'pytest-mock>=3.0 ; python_version>="3"',
+    "tox>=3.21.0",
```
tiborsimko, line 30: Actually we cannot leave it like this, since tox -e py27 would like to install tox, as part of the run, and it seems to fail for me for py27... So let's leave it out for simplicity.  Otherwise we would have to do the `python_version` dance, but it is probably not worth it.

tiborsimko (approved):

Issue #120, opened 2021-01-12
installation: xrootd troubles


tiborsimko:
Observed on a fresh vevn in Fedora 33:

```console
$  mkvirtualenv test
$ pip install '.[xrootd]' 
Processing /home/tibor/private/project/opendata/src/cernopendata-client
Collecting click>=7
  Using cached click-7.1.2-py2.py3-none-any.whl (82 kB)
Collecting requests>=2
  Using cached requests-2.25.1-py2.py3-none-any.whl (61 kB)
Collecting certifi>=2017.4.17
  Using cached certifi-2020.12.5-py2.py3-none-any.whl (147 kB)
Collecting chardet<5,>=3.0.2
  Using cached chardet-4.0.0-py2.py3-none-any.whl (178 kB)
Collecting idna<3,>=2.5
  Using cached idna-2.10-py2.py3-none-any.whl (58 kB)
Collecting urllib3<1.27,>=1.21.1
  Using cached urllib3-1.26.2-py2.py3-none-any.whl (136 kB)
Collecting xrootdpyfs>=0.2
  Using cached xrootdpyfs-0.2.1-py2.py3-none-any.whl
Collecting fs<2.0,>=0.5.4
  Using cached fs-0.5.4-py3-none-any.whl
Requirement already satisfied: setuptools in /home/tibor/.virtualenvs/test/lib/python3.9/site-packages (from fs<2.0,>=0.5.4->xrootdpyfs>=0.2->cernopendata-client==0.2.0) (51.0.0)
Collecting xrootd<5.0.0
  Using cached xrootd-4.12.6.tar.gz (2.6 MB)
ERROR: Requested xrootd<5.0.0 from https://files.pythonhosted.org/packages/26/19/d2b60f26fb539467fcbc4e4a34835d75f8c232515ba87d1c887cb0f103c5/xrootd-4.12.6.tar.gz#sha256=556f89c19c24c3ca1c0a0f9031cc575f093fd48a4538fc5e3fa9d2853e3a2a01 (from xrootdpyfs>=0.2->cernopendata-client==0.2.0) has different version in metadata: '0.0.0'
```

Note that latest stable `xrootdpyfs` needs version lower than 5.0.0, so the system one is not used...

tiborsimko:
Closed by #121.

PR #121, opened 2021-01-12
installation: pin xrootd package

tiborsimko:
Pins `xrootd` to 4.12.2 that is known to work well for compiling
latest stable `xrotdpyfs`. The later versions lead to `v-` version
detection troubles.


tiborsimko:
Note: this fixes the installation problem, but also one can see some runtime warnings:

```console
$ cernopendata-client download-files --recid 5000 --protocol xrootd
==> Downloading file 1 of 1
Plugin No such file or directory loading sec.protocol libXrdSecgsi-4.so
Plugin No such file or directory loading sec.protocol libXrdSecsss-4.so
  -> File: ./5000/2011-doubleelectron-doublemu-mueg-ttbar-1.0.0.tar.gz
==> Success!
```

Considering the file size chunking troubles and #120 I think we'll prioritise switching from xrootdpyfs to vanilla xrootd soon :wink:

```

ParthS007 (approved):
Thanks for the PR. LGTM 👍

PR #130, opened 2023-08-09
cli: add filter option for the get-metadata command

joudmas:
cli: add --filter option in get-metadata command
modified get-metadata to support multiple fields

### DIFF
--- a/cernopendata_client/cli.py
+++ b/cernopendata_client/cli.py
@@ -68,6 +68,60 @@ def version():
     display_message(msg=__version__)
 
 
+def filter_metadata(output_field, filters, output_json):
+    # noqa: D301
+    """Filter metadata objects based on specified criteria.
+
+    :param output_field: Name of the array containing objects to access
+    :param filters: Argument of the --filter option in the format some_field_name=some_value
+    :param output_json: JSON containing metadata objects
+
+    :type field: str
+    :type filters: str
+    :type output_json: list or dict
+
+    :return: None
+    """
+    filter_fields = filters.split("=")
+    if len(filter_fields) != 2:
+        display_message(
+            msg_type="error",
+            msg="Invalid filter format. Use --filter some_field_name=some_value",
+        )
+        sys.exit(1)
+    filterField_name, filterField_value = filter_fields
+    filterField_names = filterField_name.split(".")
+    for object in output_json:
+        if object == "$schema":
+            display_message(
+                msg_type="error",
+                msg="Field '{}' is not present in metadata".format(output_field),
+            )
+            sys.exit(1)
+        try:
+            if filterField_value == object[filterField_names[-1]]:
+                if output_field in object:
+                    display_message(msg=object[output_field])
+                else:
+                    display_message(msg=json.dumps(object, indent=4))
+                return
+        except (KeyError, TypeError):
+            display_message(
+                msg_type="error",
+                msg="Field '{}' is not present in metadata".format(
+                    filterField_names[-1]
+                ),
+            )
+            sys.exit(1)
+    display_message(
+        msg_type="error",
+        msg="No objects found with {}={}".format(
+            filterField_names[-1], filterField_value
+        ),
+    )
+    sys.exit(1)
+
+
 @cernopendata_client.command()
 @click.option("--recid", type=click.INT, help="Record ID")
 @click.option("--doi", help="Digital Object Identifier")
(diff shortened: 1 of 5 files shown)


cernopendata_client/cli.py
```diff
@@ -84,7 +84,14 @@ def version():
     type=click.STRING,
     help="Which CERN Open Data server to query? [default={}]".format(SERVER_HTTP_URI),
 )
-def get_metadata(server, recid, doi, title, output_value):
+@click.option(
+    "--filter",
+    "filters",
+    multiple=False,
+    help="Filter workflow that contains certain filtering criteria. "
```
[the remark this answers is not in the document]
tiborsimko: Moreover, you can enrich the user guide `usage.rst` the section "Getting metadata" so that the David Rousseau example will also appear here: https://cernopendata-client.readthedocs.io/en/latest/usage.html#getting-metadata

docs/usage.rst
```diff
@@ -87,6 +87,30 @@ datasets, you can use **--output-value** command-line option:
     $ cernopendata-client get-metadata --recid 1 --output-value system_details.global_tag
     FT_R_42_V10A::All
 
+If you would like to extract parts of the metadata by filtering some
```
tiborsimko, line 90: You can rather say:

- If the output field produces a list of values, you may want to filter only certain field values of interest. For example, you may want to ask not to output all authors of a dataset record, but only the authors matching certain particular author name. You can use the `--filter` command-line option to achieve this:

docs/usage.rst
```diff
@@ -87,6 +87,30 @@ datasets, you can use **--output-value** command-line option:
...
+.. code-block:: console
+
+    $ cernopendata-client get-metadata --recid 329 --output-value authors.orcid
+    0000-0002-2634-4958
+    0000-0001-8363-9827
+    0000-0002-9764-9783
+    0000-0002-9266-1783
+    0000-0001-5610-1060
+    0000-0001-7613-8063
+    $ cernopendata-client get-metadata --recid 329 --output-value authors --filter name='Rousseau, David'
```
tiborsimko, line 104: This one comes "in the middle" of the ORCID example. I think we have two options:

1. First show example getting all authors, and filtering on one. Then show another example how to get only ORCID identifiers.
2. Alternatively, we don't show full author JSON output, only the ORCID filtering.

tests/test_cli_get_metadata.py
```diff
@@ -88,6 +115,24 @@ def test_get_metadata_from_output_fields_wrong():
...
+    test_get_metadata = CliRunner()
+    test_result = test_get_metadata.invoke(
+        get_metadata,
+        [
+            "--recid",
+            329,
+            "--output-value",
+            "authors.orcid",
+            "--filter",
+            "authors.url=Rousseau, David",
```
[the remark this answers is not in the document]
tiborsimko: This works well with the PR update :+1: You may want to also add a test case for this, i.e. get all authors, then with the filter, and then compare the two lists whether the second one has one less item as expected

cernopendata_client/cli.py
```diff
@@ -84,7 +84,14 @@ def version():
     type=click.STRING,
     help="Which CERN Open Data server to query? [default={}]".format(SERVER_HTTP_URI),
 )
-def get_metadata(server, recid, doi, title, output_value):
+@click.option(
+    "--filter",
+    "filters",
+    multiple=False,
```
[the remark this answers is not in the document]
joudmas: I will create a new issue for this and link it back to this discussion for reference.

cernopendata_client/cli.py
```diff
@@ -68,6 +68,60 @@ def version():
     display_message(msg=__version__)
 
 
+def filter_metadata(output_field, filters, output_json):
+    # noqa: D301
+    """Filter metadata objects based on specified criteria.
+
+    :param field: Name of the array containing objects to access
```
tiborsimko, line 75: If the first argument is named `output_field`, then we should also update the docstring of the function, see the lines:

```
:param field:

...

:type field:
```

cernopendata_client/cli.py
```diff
@@ -68,6 +68,60 @@ def version():
     display_message(msg=__version__)
 
 
+def filter_metadata(output_field, filters, output_json):
+    # noqa: D301
+    """Filter metadata objects based on specified criteria.
+
+    :param output_field: Name of the array containing objects to access
+    :param filters: Argument of the --filter option in the format field_name=value
```
tiborsimko, line 76: `field_name=value` could be misunderstood: is value supposed to be the field name? Or the value of the field?  Therefore it would be better to keep "somefield=somevalue" speaking abstractly, or say "some_field_name=some_field_value" to be more precise.

tiborsimko (approved):
Thanks, works nicely :+1: 

We can merge this initial version, and please create that additional issue for multiple filters.

PR #132, opened 2023-08-09
codecov: add threshold values

tiborsimko:
Adds desired targets and threshold values for code coverage tests in pull requests.

Removes custom code coverage range to fall back to codecov defaults.


mdonadoni (approved):
LGTM!

PR #135, opened 2023-10-30
docker: optimise image and change base Fedora 38

tiborsimko:
Upgrades the base container image to Fedora 38 and optimises the container image building process by pre-installing system Python packages. This helps (i) time-wise, because lengthy XRootD package installation during `pip` is no longer necessary; (ii) space-wise, because several unnecessary compilation-time dependencies are not installed anymore. The size of the produced container image is reduced to one half.


DaanRosendal (approved):
LGTM! Built and ran docker image without issues.

PR #157, opened 2025-10-05
test(verifier): replace `sed` calls with native Python code

tiborsimko:
On macOS, `sed -i -e` creates backup files with `-e` suffix, causing test failures due to unexpected files in the test directory. This commit replaces `sed` commands with native Python file operations for better platform independence of tests.


psaiz (approved):

michaelbuchar:
Moreover, are you planning on solving the failing CI checks? I believe the failure I got referenced in the above comment may be related to that.

PR #158, opened 2025-10-05
build(python): add support for Python 3.14 (#158)

tiborsimko:


michaelbuchar (approved):
Runs like a charm on my end as well! 🪄

PR #176, opened 2026-01-25
Explicitly close files to prevent resource leak

Stephen0512:
This pull request fixes a file-handling issue in the `verifier`. The current implementation uses `open()` without explicitly closing the file, which can potentially lead to resource leaks. While Python's garbage collector will eventually close the file once it is no longer referenced, it is generally better practice to explicitly close files or use a context manager.

This change replaces the direct `open()` call with a `with` statement to ensure proper file closure and follow Python best practices. The issue was identified during an ongoing research project.

### DIFF
--- a/cernopendata_client/verifier.py
+++ b/cernopendata_client/verifier.py
@@ -2,7 +2,7 @@
 #
 # This file is part of cernopendata-client.
 #
-# Copyright (C) 2020, 2025 CERN.
+# Copyright (C) 2020, 2025, 2026 CERN.
 #
 # cernopendata-client is free software; you can redistribute it and/or modify
 # it under the terms of the GPLv3 license; see LICENSE file for more details.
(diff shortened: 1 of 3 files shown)


cernopendata_client/verifier.py
```diff
@@ -38,9 +38,8 @@ def get_file_checksum(afile):
     :return: Adler32 checksum of file
     :rtype: str
     """
-    return "adler32:{:08x}".format(
-        zlib.adler32(open(afile, "rb").read(), 1) & 0xFFFFFFFF
-    )
+    with open(afile, "rb") as f:
+        return "adler32:{:08x}".format(zlib.adler32(f.read(), 1) & 0xFFFFFFFF)
```
tiborsimko, line 42: @Stephen0512 Thanks for the contribution! I have simplified this into one line, added you to the list of co-authors, rebased against latest `master` branch, and rephrased commit message to suit our Release-Please based practices.

tiborsimko (approved):

PR #182, opened 2026-06-17
fix(walker): report concrete XRootD client errors

tiborsimko:
Stop masking XRootD failures behind generic download and directory
messages. Check the status objects returned by CopyProcess.prepare(),
CopyProcess.run(), and FileSystem.dirlist(), and surface the underlying
XRootD message when the operation fails.

Keep the existing user-facing "directory does not exist" message for
missing paths by recognising both the XRootD client not-found status and
the server-side kXR_NotFound errno.

Note that a failed XRootD transfer now aborts the command with exit
code 1 instead of continuing to the next file, which could previously
end in an unhandled error or a misleading "Success!".

Detect traversable directory entries using the named XRootD X_BIT_SET,
IS_DIR, and IS_READABLE flags instead of comparing the complete flag
value to numeric mask 19. This skips inaccessible directories while
allowing readable directories with additional flags to be traversed.
Append each listed path only once.

### DIFF
--- a/cernopendata_client/downloader.py
+++ b/cernopendata_client/downloader.py
@@ -180,24 +180,26 @@ class DownloaderXrootd:
 
     def file_downloader(self):
         """Download single file with XRootD."""
-        try:
+        display_message(
+            msg_type="note",
+            msg="File: ./{}/{}".format(
+                self.path,
+                self.file_name,
+            ),
+        )
+        process = xrootdclient.CopyProcess()
+        process.add_job(
+            SERVER_ROOT_URI + self.file_src, os.getcwd() + os.sep + self.file_dest
+        )
+        status = process.prepare()
+        if status.ok:
+            status, _ = process.run()
+        if not status.ok:
             display_message(
-                msg_type="note",
-                msg="File: ./{}/{}".format(
-                    self.path,
-                    self.file_name,
-                ),
+                msg_type="error",
+                msg="Download error: {}".format(status.message.strip()),
             )
-            process = xrootdclient.CopyProcess()
-            process.add_job(
-                SERVER_ROOT_URI + self.file_src, os.getcwd() + os.sep + self.file_dest
-            )
-            process.prepare()
-            process.run()
-        except Exception:
-            display_message(
-                msg_type="error", msg="Download error occured. Please try again."
-            )
+            sys.exit(1)
 
 
 def check_error(
--- a/cernopendata_client/walker.py
+++ b/cernopendata_client/walker.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 # This file is part of cernopendata-client.
 #
-# Copyright (C) 2020, 2021 CERN.
+# Copyright (C) 2020, 2021, 2026 CERN.
 #
 # cernopendata-client is free software; you can redistribute it and/or modify
 # it under the terms of the GPLv3 license; see LICENSE file for more details.
@@ -17,14 +17,24 @@ import datetime
 from .config import SERVER_ROOT_URI
 from .printer import display_message
 
+# XRootD client error code for a not-found path. Resolved at import time and
+# left as None for older bindings that do not expose it; the wire-protocol
+# errno below is the stable fallback.
+XROOTD_CODE_NOT_FOUND = None
+
 try:
     from XRootD import client as xrootdclient
-    from XRootD.client.flags import DirListFlags
+    from XRootD.client.flags import DirListFlags, StatInfoFlags
+    from XRootD.client.responses import XRootDStatus
 
+    XROOTD_CODE_NOT_FOUND = getattr(XRootDStatus, "errNotFound", None)
     FS = xrootdclient.FileSystem(SERVER_ROOT_URI)
     xrootd_available = True
 except ImportError:
     xrootd_available = False
+
+# XRootD server error number returned when a path does not exist (kXR_NotFound).
+XROOTD_ERRNO_NOT_FOUND = 3011
 
 
 def get_list_directory(path, recursive, timeout, time_start=None):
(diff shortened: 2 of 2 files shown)


cernopendata_client/downloader.py
```diff
@@ -180,24 +180,26 @@ def show_download_progress(self):
...
+                self.file_name,
+            ),
+        )
+        process = xrootdclient.CopyProcess()
+        process.add_job(
+            SERVER_ROOT_URI + self.file_src, os.getcwd() + os.sep + self.file_dest
+        )
+        status = process.prepare()
+        if status.ok:
+            status, _ = process.run()
```
psaiz, line 196: What is the second return value from run? Is it the error, similar to status.message?

cernopendata_client/downloader.py
```diff
@@ -180,24 +180,26 @@ def show_download_progress(self):
...
-        except Exception:
+        display_message(
+            msg_type="note",
+            msg="File: ./{}/{}".format(
+                self.path,
+                self.file_name,
+            ),
+        )
+        process = xrootdclient.CopyProcess()
+        process.add_job(
```
psaiz, line 191: could this one ( or line 190) fail? If so, is it with an exception?

cernopendata_client/downloader.py
```diff
@@ -180,24 +180,26 @@ def show_download_progress(self):
...
+        status = process.prepare()
+        if status.ok:
+            status, _ = process.run()
+        if not status.ok:
             display_message(
-                msg_type="error", msg="Download error occured. Please try again."
+                msg_type="error",
+                msg="Download error: {}".format(status.message.strip()),
             )
+            sys.exit(1)
```
psaiz, line 202: Instead of exit, wouldn't it be better to propagate the error to the caller?

cernopendata_client/walker.py
```diff
@@ -57,24 +67,33 @@ def get_list_directory(path, recursive, timeout, time_start=None):
...
+            display_message(
+                msg_type="error",
+                msg="Directory {} does not exist.".format(path),
+            )
+        else:
+            display_message(
+                msg_type="error",
+                msg="Cannot list directory {}: {}".format(path, status.message.strip()),
+            )
         sys.exit(1)
```
psaiz, line 87: Same thing as above with the sys.exit. And, since this might be a recursive function, it might be even more important to pass the error instead of just exiting

cernopendata_client/walker.py
```diff
@@ -57,24 +67,33 @@ def get_list_directory(path, recursive, timeout, time_start=None):
...
+    for entry in listing:
+        if entry.statinfo.flags == 19:  # entry is a directory
+            files.append(path + os.sep + entry.name)
```
psaiz: This line is in both sides of the if/else. What about doing it outside the if/else?

cernopendata_client/walker.py
```diff
@@ -57,24 +67,33 @@ def get_list_directory(path, recursive, timeout, time_start=None):
...
+                msg="Directory {} does not exist.".format(path),
+            )
+        else:
+            display_message(
+                msg_type="error",
+                msg="Cannot list directory {}: {}".format(path, status.message.strip()),
+            )
         sys.exit(1)
+    for entry in listing:
+        if entry.statinfo.flags == 19:  # entry is a directory
```
psaiz: Magic number 19. From where does it appear?

tiborsimko: This predates the PR. It is the XRootD stat bitmask `X_BIT_SET (1) | IS_DIR (2) | IS_READABLE (16)`. I agree that both the magic number and exact equality are brittle: a directory with additional or different permission bits can be misclassified as a file. I will import `StatInfoFlags` inside the existing optional-XRootD import block and test `entry.statinfo.flags & StatInfoFlags.IS_DIR`. Keeping that import inside the guarded block preserves normal client use on systems where XRootD is not installed.

cernopendata_client/walker.py
tiborsimko: Agreed. Both branches append the same path. I will compute `entry_path`, append it once before the conditional, and use that value for recursive calls. This preserves the result order and combines directly with the directory-flag correction below.

cernopendata_client/walker.py
tiborsimko: The previous implementation already called `sys.exit(1)` from its exception handler; this patch changes how the XRootD failure is detected but preserves that behavior. Since `sys.exit()` raises `SystemExit`, it immediately unwinds every recursive frame and cannot return a partial result. A clean exception refactor would need to cover the XRootD-unavailable and timeout exits too, preserve their different exit codes, and translate the exception once in `list_directory()`. I propose keeping this PR focused and handling such an API-level refactor separately if desired.

cernopendata_client/downloader.py
tiborsimko: This `sys.exit(1)` is a deliberate behavior change. Previously, an operational XRootD copy failure returned a non-OK status rather than raising, so the broad `except Exception` did not handle it. The caller then ran `check_error()` unconditionally: if no destination file had been created, `os.path.getsize()` raised an unhandled `FileNotFoundError`; if a partial file existed, the command could continue to later files and could eventually print `Success!`. The new behavior aborts on the first failed XRootD transfer and reports the concrete XRootD error instead. `sys.exit()` raises `SystemExit`, so the failure propagates through `download_single_file()` with the intended CLI exit code and without a traceback. I agree that a domain exception caught at the CLI boundary would be cleaner for a reusable API, but that should be applied consistently across the downloader engines rather than only this branch.

cernopendata_client/downloader.py
tiborsimko: They can theoretically raise a Python exception for binding misuse or an internal binding problem. Expected operational copy failures are reported through the statuses returned by `prepare()` and `run()`; the Python `add_job()` wrapper does not expose its underlying status return. I removed the broad `except Exception` intentionally so an unexpected programming or binding failure is not relabelled as a generic transfer failure. With the known string arguments used here, ordinary transfer failures follow the status path.

cernopendata_client/downloader.py
tiborsimko: `run()` returns `(status, results)`. The second value is a list of per-job result dictionaries; each dictionary may include its own `status` and copy metadata. It is not another error string. This `CopyProcess` contains exactly one job, and XRootD's overall status carries that job's failure, so `status.message` is the error we want to display. The `_` intentionally discards per-job details that this caller does not otherwise use.

cernopendata_client/walker.py
tiborsimko: Code amended.

cernopendata_client/walker.py
tiborsimko: Code amended.

cernopendata_client/walker.py
tiborsimko: Follow-up: Docker CI test exposed an unreadable EOS directory whose flags contain only IS_DIR (2). The final implementation therefore requires X_BIT_SET, IS_DIR, and IS_READABLE, while allowing additional flags, instead of testing IS_DIR alone or comparing the complete value with numeric mask 19. This preserves the original traversability requirements and avoids recursing into inaccessible directories. Code amended, and Docker CI test now passes.

psaiz (approved):

PR #183, opened 2026-08-06
fix(downloader): reject incomplete HTTP downloads

### LINKED ISSUE #140: downloader: partial downloads are marked as success:
```
[psaiz@aiadm81 ~]$ cernopendata-client download-files --filter-range 1 --recid 14
==> Downloading file 1 of 1
  -> File 00459D48-EB70-E011-AF09-90E6BA19A252.root is incomplete. Resuming download.
  -> File: ./14/00459D48-EB70-E011-AF09-90E6BA19A252.root
  -> Progress: 61758/1486456 KiB (4%)
==> Success!
[psaiz@aiadm81 ~]$ ls 14/00459D48-EB70-E011-AF09-90E6BA19A252.root -al
-rw-r--r--. 1 psaiz z2 63240651 Nov 20 11:56 14/00459D48-EB70-E011-AF09-90E6BA19A252.root
[psaiz@aiadm81 ~]$ cernopendata-client download-files --filter-range 1 --recid 14
==> Downloading file 1 of 1
  -> File 00459D48-EB70-E011-AF09-90E6BA19A252.root is incomplete. Resuming download.
  -> File: ./14/00459D48-EB70-E011-AF09-90E6BA19A252.root
  -> Progress: 93190/1486456 KiB (6%)
==> Success!
[psaiz@aiadm81 ~]$ ls 14/00459D48-EB70-E011-AF09-90E6BA19A252.root -altdh
-rw-r--r--. 1 psaiz z2 92M Nov 20 11:57 14/00459D48-EB70-E011-AF09-90E6BA19A252.root
[psaiz@aiadm81 ~]$ 
```

The message of succes should not be there.

divyanshu-iitian:
## Summary

- re-check HTTP file size after the selected downloader returns
- exit with an explicit incomplete-file error instead of allowing the CLI to print `Success!`
- add a local regression test for an interrupted resumed download

This keeps the existing resume behavior, but closes the gap where an HTTP transfer could stop early without raising from the download engine. The final HEAD/local-size comparison now prevents a partial scientific data file from being reported as complete.

Closes #140.

## Validation

- `pytest -m local tests/test_downloader.py -q` (`18 passed`)
- local suite excluding the two line-ending-sensitive checksum fixtures: `40 passed`
- `black --check cernopendata_client tests`
- `flake8 cernopendata_client tests`
- `pydocstyle cernopendata_client tests`


tiborsimko:
Thank you for spotting this problem and proposing a fix. A truncated download must not finish with a successful exit status.

We already have `--verify`, which is intended to check downloaded files against the size and Adler-32 checksum in the record metadata. `get_files_list()` already returns these authoritative values before the download starts, so the client does not need to ask the server for the size again. Rather than add another, HTTP-only verification layer based on an extra `HEAD` request, I think we should fix the current limitations of `--verify` and enable metadata-based verification by default. The current `HEAD` error path can also leave `file_size_online` undefined when the request fails.

I have opened #185 to track that work. In favour of that single verification path, I would close this PR. Thank you again for highlighting the underlying correctness issue.

PR #186, opened 2026-08-14
feat(download): verify downloaded files by default

### LINKED ISSUE #185: RFC verify downloaded files by default
## Problem

`download-files` can currently report success after an incomplete transfer.
The existing `--verify` option is intended to compare downloaded files with the
size and Adler-32 checksum in the record metadata, but it is opt-in and has
correctness and performance limitations that must be fixed before it can safely
become the default.

We should use the record metadata already fetched by the client instead of
adding a separate HTTP `HEAD`-based check.

Related download and resume failures should also produce clear errors rather
than tracebacks or successful exit statuses.

## Proposed behavior

- Always compare each exact downloaded destination with its expected metadata
  size.
- Verify its metadata checksum by default when a checksum is available.
- Provide `--verify/--no-verify`, defaulting to verification; `--no-verify`
  skips only checksum calculation, not the basic size check.
- Support direct files, file-index entries, filtered downloads, duplicate
  basenames stored in subdirectories, and all download engines consistently.
- Calculate checksums with bounded memory and avoid repeated metadata queries or
  repeated verification of files already checked.
- Report transfer and verification failures with a nonzero exit status, and
  print `Success!` only after all requested files pass.

## Acceptance criteria

- Incomplete, oversized, or checksum-mismatched downloads cannot report
  success.
- Verification uses the exact local destination and the metadata already
  obtained for that download.
- Default verification works for HTTP(S), XRootD, expanded file indices,
  filters, and duplicate filenames. Unexpanded file-index responses are either
  validated against applicable metadata or explicitly excluded with the reason
  documented.
- `--no-verify` retains the expected-size check.
- Checksum calculation is incremental, and verification work is linear in the
  downloaded data size.
- CLI help, documentation, tests, and release notes describe the new default.

This issue supersedes the additional post-download `HEAD` check proposed in
PR #183.

tiborsimko:
Carry record metadata through file selection and destination mapping so
each download is checked against its exact local path without extra
metadata or HEAD requests. Always reject missing, incomplete, and
oversized files. Verify Adler-32 checksums by default while allowing
--no-verify to skip only the checksum pass.

Cover direct files, expanded file indices, filters, duplicate
basenames, and all download engines through the same verification path.
Derive duplicate-basename layouts from the complete record even for
filtered selections so successive downloads cannot overwrite each other.
Explicitly skip size and checksum comparisons for unexpanded index
responses because their metadata size describes the aggregate indexed
data rather than the JSON document.

Calculate checksums incrementally and make standalone verification
recursive. Retain the selected engine during known-error-page retries.
Document the new default and add local and live regression coverage.

Closes #185

### DIFF
--- a/cernopendata_client/cli.py
+++ b/cernopendata_client/cli.py
@@ -16,7 +16,7 @@ import sys
 import re
 
 from .searcher import (
-    get_file_info_remote,
+    get_file_entries,
     get_files_list,
     get_recid,
     get_recid_api,
@@ -29,8 +29,7 @@ from .downloader import (
     get_download_files_by_name,
     get_download_files_by_range,
     get_download_files_by_regexp,
-    get_download_path,
-    get_file_subdirectories,
+    get_download_items,
 )
 from .validator import (
     validate_range,
@@ -41,7 +40,7 @@ from .validator import (
     validate_retry_sleep,
 )
 from .walker import get_list_directory
-from .verifier import get_file_info_local, verify_file_info
+from .verifier import get_local_file_paths, verify_downloaded_file
 from .metadater import filter_metadata, handle_error_message
 from .config import (
     SERVER_HTTP_URI,
@@ -245,11 +244,10 @@ def get_file_locations(server, recid, doi, title, protocol, expand, verbose):
     help="Download files from a specified list range (i-j)",
 )
 @click.option(
-    "--verify",
+    "--verify/--no-verify",
     "verify",
-    is_flag=True,
-    default=False,
-    help="Verify downloaded data file integrity",
+    default=True,
+    help="Verify downloaded data file checksums when available [default=yes]",
 )
 @click.option(
     "--retry-limit",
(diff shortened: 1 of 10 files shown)


cernopendata_client/cli.py
```diff
@@ -245,11 +244,10 @@ def get_file_locations(server, recid, doi, title, protocol, expand, verbose):
     help="Download files from a specified list range (i-j)",
 )
 @click.option(
-    "--verify",
+    "--verify/--no-verify",
     "verify",
-    is_flag=True,
```
psaiz, line 250: Why is this not a flag?

cernopendata_client/cli.py
```diff
@@ -437,12 +448,17 @@ def verify_files(server, recid, doi, title):
     record_json = get_record_as_json(server, recid, doi, title)
     record_recid = record_json["metadata"]["recid"]
```
psaiz, line 449: Later on, the code uses the string version of this. What about converting it to str already here?

cernopendata_client/cli.py
```diff
@@ -458,17 +474,22 @@ def verify_files(server, recid, doi, title):
...
         display_message(
             msg_type="error",
             msg="File count does not match.",
         )
         sys.exit(1)
 
-    # Verify size and checksum of each file
-    verify_file_info(file_info_local, file_info_remote)
+    # Verify size and checksum of each exact destination
+    for download_item in download_items:
```
psaiz, line 487: This will check the list of download_items. What happens if local_file_paths give a different number of entries than the download_items?

cernopendata_client/searcher.py
```diff
@@ -13,12 +13,23 @@
...
 
 from .config import SERVER_HTTP_URI, SERVER_ROOT_URI, SERVER_HTTPS_URI
 from .printer import display_message
 
 
+@dataclass(frozen=True)
+class FileEntry:
+    """Describe a downloadable file and its verification metadata."""
+
+    uri: str
```
psaiz, line 27: Would it make sense to put another attribute to identify if a file is only in the cold storage?

